PR #2, opened 2022-02-21
UI Probe for node instead of nginx

gildub:
UI is not using Nginx anymore and the deployment will constantly kill the ui pod because the probe would fail.


fbladilo (approved):
LGTM

Issue #14, opened 2022-02-28
Install windup addon.


jortel:
The operator needs to install the windup addon.  This includes:

The windup CR in the tackle namespace.
```
---
kind: Addon
apiVersion: tackle.konveyor.io/v1alpha1
metadata:
  name: windup
spec:
  image: quay.io/konveyor/tackle2-addon-windup.
  mounts:
  - claim: m2
    name: m2
```

The claim is RWX and can follow the same pattern as buckets.  The same initial size is fine for now.  We can properly adjust the default before GA.

Related to: https://github.com/konveyor/tackle2-operator/issues/13

fbladilo:
Closed , resolved via #15

PR #15, opened 2022-03-01
Create addon windup CR and other fixes

### LINKED ISSUE #13: Easier support for RWX volumes for buckets.
Currently, the operator supports customization for RWX volumes for buckets primarily with the `hub_bucket_volume_claim_name` setting. This approach requires user to create both the PV and PVC by hand.  I wonder if it would be a good improvement to add another variable `hub_rwx_storage_class` that can be blank by default but used to create the bucket PVC when specified.  For example, to use NFS, the user can create a PV with a storage class of "nfs".  Then, when installing tackle, the user creates the CR with:
```
hub_rwx_storage_class: nfs
```
and the operator does the rest.  Just 1 less thing for the user to do.

fbladilo:
This PR does the following : 

* Create the addon CR in Tackle namespace and corresponding PVC
* Addon CR reference to windup-addon quay.io repository and PVC name , the m2 reference is fixed , not configurable via operator.
* Implement tackle_rwx_storage_class and tackle_rwo_storage_class
       -  tackle_rwx_storage_class will set the storageclass for all Tackle RWX requested PVCs
       -  tackle_rwo_storage_class will set the storageclass for all Tackle RWO requested PVCs
* Update RBAC to allow operator SA to handle addons
* Update documentation with important operator settings, storage requirements and other minor fixes
* Fixes #13 and #14 
* Tested on OCP v4.10 with NFS PVs


README.md
```diff
@@ -13,26 +20,69 @@ Installing latest requires creating a new catalog source.
...
+
+Name | Default | Description
+--- | --- | ---
+hub_database_volume_size | 5Gi | Size requested for Hub database volume
+hub_bucket_volume_size | 100gi | Size requested for Hub bucket volume
+keycloak_database_data_volume_size | 1Gi | Size requested for Keycloak DB volume
+pathfinder_database_data_volume_size | 1Gi | Size requested for Pathfinder DB volume
+windup_data_volume_size | 100Gi | Size requested for Windup addon volume
+tackle_rwx_storage_class | N/A | Storage class requested for Tackle RWX volumes
+tackle_rwo_storage_class | N/A | Storage class requested for Tackle RWO volumes
```
jortel, line 51: Since all of these variables are `tackle` operator settings, I wonder if tackle_ is useful.

README.md
```diff
@@ -13,26 +20,69 @@ Installing latest requires creating a new catalog source.
...
-## Customize Settings
+If operator defaults need to be altered, the Tackle CR spec can be customized to meet desired needs, see the table below for some of the most significant settings:
+
+Name | Default | Description
+--- | --- | ---
+hub_database_volume_size | 5Gi | Size requested for Hub database volume
+hub_bucket_volume_size | 100gi | Size requested for Hub bucket volume
+keycloak_database_data_volume_size | 1Gi | Size requested for Keycloak DB volume
+pathfinder_database_data_volume_size | 1Gi | Size requested for Pathfinder DB volume
+windup_data_volume_size | 100Gi | Size requested for Windup addon volume
```
jortel, line 49: Seems better for the description to match the storage requirements below:
`Windup maven m2 repository`

README.md
```diff
@@ -13,26 +20,69 @@ Installing latest requires creating a new catalog source.
...
+
+```
+kind: Tackle
+apiVersion: tackle.konveyor.io/v1alpha1
+metadata:
+  name: tackle
+  namespace: konveyor-tackle
+spec:
+  hub_bucket_volume_size: "50Gi"
+  tackle_rwx_storage_class: ""
```
jortel, line 83: On the assumption we agreed on PV with "" storage classes were edge cases, would it be better for the example to be, say ... tackle_rwx_storage_class: "nfs"?

jortel:
Few comments but LTGM.
Thanks!

README.md
fbladilo: My rationale was for the tackle_ naming was to hint that the variables affected *all* RWX or RWO volumes, to answer the question specifically no I don't think we will ever see non-tackle related variables.

README.md
fbladilo: Sure , corrected.

README.md
fbladilo: corrected

README.md
jortel: Don't mean to be pedantic but I think the tackle_ prefix is unnecessary and mildly suggests that the other variables may not be tackle related.

fbladilo: Removed unnecessary tackle_ prefixing to storage class variables

PR #19, opened 2022-03-03
Switch PG Docker Hub image to RH quay based

fbladilo:
The purpose of this PR is deprecate the use of public Docker Hub PG images in favor of RH SCL based supported images which align better with the RH ecosystem and will reduce integration friction. The PG version has not been changed and remains in v12.


jortel (approved):
LGTM

PR #31, opened 2022-03-22
Add tackle-hub client and roles to realm and configure hub for auth

mansam:
* Added tackle-hub client to tackle realm.
* Added scopes for each resource collection and method.
* Added tackle-admin, tackle-architect, and tackle-migrator roles.
* Set up Keycloak env variables in Hub deployment


fbladilo (approved):

fbladilo:
@mansam LGTM

PR #32, opened 2022-03-23
Add hub route/ingress

### LINKED ISSUE #28: Add route to the hub API.
Add route to the hub API: hub-konveyor-tackle.apps.cluster-jortel.v2v.bos.redhat.com with TLS.

We may want to rename the UI route to reduce the number of "tackle" in it and be more specific.
Suggest:  ui-konveyor-tackle

I'm not sure what the convention is but assume it is to include "konveyor" and "tackle" in it somewhere.

fbladilo:
- Fixes #28


fbladilo:
Closing as this is not needed any longer

PR #40, opened 2022-03-31
Realm improvements

mansam:
* Add `admin` user with a temporary password that has `admin` (used by Pathfinder) and `tackle-admin` roles.
* Remove write permissions on proxies/identities/settings from the `tackle-architect` role.
* Add `adoptionplans:post` scope to all tackle roles.


jortel:
Can you document the admin user and temporary password in the README?

PR #50, opened 2022-04-28
Keycloak: Use init container image instead of busybox

fbladilo:
Currently we use a busybox image which calls wget to obtain theme .jar file and inject it into a volume for keycloak use , this new image based approach does the same but has the advantage in proxy configurations where sites might not allow to download files directly , only images.


jortel (approved):
LGTM

PR #55, opened 2022-05-05
Specify Keycloak User password format

aufi:
Adding password format requirement to the Keycloak initial Tackle realm configuration.

Proposal to a valid password format is to:
- have minimal length 8 chars
- contain at least 1 uppercase letter
- contain at least 1 lowercase letter
- contain at least 1 digit

**The password format could be discussed in reviews.**

More possible options are in Keycloak docs:https://github.com/keycloak/keycloak-documentation/blob/main/server_admin/topics/authentication/password-policies.adoc

https://issues.redhat.com/browse/TACKLE-522


jortel (approved):
LGTM

PR #58, opened 2022-05-09
WIP: Initial commit for keycloak v18 integration

fbladilo:


fbladilo:
Closing, no need to ugprade keycloak in the near term , will re-open/re-assess when necessary

PR #62, opened 2022-05-10
Wait for PG deployments to be in running state

fbladilo:
- Allow PG pods to be in running state before continuing deployment of
  keycloak and pathfinder
- Fixes race issues when PG is not yet forwarding traffic while other
  applications are trying to access during start-up

### DIFF
--- a/roles/tackle/tasks/main.yml
+++ b/roles/tackle/tasks/main.yml
@@ -64,6 +64,19 @@
           state: present
           definition: "{{ lookup('template', 'deployment-keycloak-postgresql.yml.j2') }}"
 
+      - name: "Check status of Keycloak PostgreSQL Deployment"
+        k8s_info:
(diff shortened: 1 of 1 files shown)


jortel (approved):
LGTM

roles/tackle/tasks/main.yml
```diff
@@ -64,6 +64,19 @@
...
+        k8s_info:
+          api_version: v1
+          kind: Pod
+          namespace: "{{ app_namespace }}"
+          label_selectors:
+             - app.kubernetes.io/name = {{ keycloak_database_service_name }}
+        register: pod
+        until: pod.resources[0].status.phase == "Running"
+        retries: 4
+        delay: 10
```
jortel, line 77: Let's consider retries:20 and delay:2 for smoother, quicker feel.

PR #78, opened 2022-05-23
Add entry for exposing upload size to the ui pod via env var

ibolton336:
- Loads an env var into the ui pod for determining file upload size. Helps to solve https://issues.redhat.com/browse/TACKLE-575


fbladilo (approved):
LGTM

PR #79, opened 2022-05-23
Allow tackle-hub to use anyuid scc

### LINKED ISSUE #76: Hub SA security context.
Update tackle-hub SA like:
```
oc adm policy add-scc-to-user anyuid -z tackle-hub
```
Probably need explict namespace instead of `-z`.

Needed so hub can run task pods with explicit security context to:
```
runAsUser: 0 (actual root)
```
This is needed so addon SSH component can create `/etc/ssh/known_hosts`  Also, make troubleshooting MUCH easier when ssh(ing) to the pod.  The injected (generated) user on OCP is a pain.

This is mainly need for OCP.

fbladilo:
- Fixes #76

### DIFF
--- a/bundle/manifests/tackle-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/tackle-operator.clusterserviceversion.yaml
@@ -261,6 +261,14 @@ spec:
           - '*'
           verbs:
           - '*'
+        - apiGroups:
+          - security.openshift.io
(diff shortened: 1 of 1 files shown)


jortel (approved):
LGTM
Tested on my PSI cluster and hub able to create the pod with securityContext and runAsUser.

PR #82, opened 2022-05-26
Allow namespace isolation via network policies

fbladilo:
- Implement namespace isolation via network policies, by default feature is turned on. OCP has support for network policies by default in OpenShift SDN , in k8s clusters calico can be installed
- Policy allows ingress traffic to all ports from any pod that lives in the Tackle namespace, blocks all other ingress traffic
- Enforcement of policy via immutable namespace label kubernetes.io/metadata.name=<namespace>
- If the cluster lacks support for k8s network policies , the policy will not have any effect
- Bumped minKubeVersion to v1.22.0 for network policy feature compatibility
- Tested on OCP v4.10 and minikube v1.24 w/ calico


jortel (approved):
LGTM

PR #86, opened 2022-05-27
Allow external UI access via network policy

fbladilo:
**Network policy updates**

* tackle-default-deny : explicitly denies all ingress traffic to all pods in tackle namespace from any source
* tackle-namespace : allows ingress traffic from any pod in the tackle namespace to all ports
* tackle-external : allows ingress traffic from any pod to the tackle UI pod on ports 8080 and 8443

I have left the feature namespace isolation off by default until the policy gets more mileage.


jortel (approved):
LGTM

PR #91, opened 2022-06-06
Fix operator details page in OCP web console

fbladilo:
- Incorrect API group causes OCP web console to fail to render with an
  error message


jortel (approved):
LGTM

Issue #101, opened 2022-07-20
Add flag to Tackle CR to change the require password setting


wkulhanek:
Currently when setting up a new Tackle 2 instance via the Operator the Keycloak configuration requires a change of password on the first login. This prevents automatic seeding of Tackle for demo / lab / workshop environments - unless authentication is turned off. Which is not desirable.

Add a flag to the Tackle CR to optionally turn off the requirement to change the password. Which would change the configuration config map (https://github.com/konveyor/tackle2-operator/blob/main/roles/tackle/templates/configmap-keycloak-sso.yml.j2) - guessing around line 530.

Also potentially add a field to the Tackle CR to specify a password. Again this could be generated during automatic deployment and would not be less secure than the current default 'password'. 

@rromannissen @aufi FYI

fbladilo:
This has been addressed by #105 , closing.

PR #105, opened 2022-07-22
Keycloak allow control of password update requirement

fbladilo:
- See #101
- Adds keycloak_sso_req_passwd_update boolean to control password update behavior via Tackle CR (by default enabled)
- Since keycloak imports the config once, variable must be present in Tackle CR on first deployment to take effect


jortel (approved):
LGTM

PR #108, opened 2022-08-09
Rename component image ENV vars

fbladilo:
- Rename image environmental vars to comply with OSBS SHA pinning naming scheme
- Requires re-deployment as operator and OLM metadata has changed


jortel (approved):
LGTM

PR #111, opened 2022-09-02
ci(actions): start minikube & install tackle

djzager:
This is the first PR with the goal leveraging GitHub Actions to provide the Konveyor organization with a simple and extensible CI/CD platform. Here is a [link to the Google Slides](https://docs.google.com/presentation/d/1l9BObaU7tSO9OXQw5CCs_ziyJIexw6qxZi-o2wRtNVc) where this was presented in the [Konveyor Technical Architecture Forum](https://docs.google.com/document/d/1Z_i9Qzxn8GST5b6aU_-IUBY4fcd13YCwCsfEqOnAff8/edit#heading=h.ddgaq7lqxzdk).

**tl;dr**

* This PR introduces 2 GitHub Actions (`start-minikube` and `install-tackle`)
* These actions are run via shell scripts that can be invoked via Makefile for local development
* Shell scripts are intentionally made to be self-contained (no `source utils.sh`) and idempotent (`kubectl apply` instead of `kubectl create`) to make them useful for local development
* These actions (and shell scripts) can be re-used in other Tackle projects that need them for testing (ie. `tackle2-addon-windup`)

The end goal is to make these scripts useful in the typical development workflow to simplify onboarding new contributors by having well used/tested startup scripts.

### DIFF
--- a/.github/actions/install-tackle/action.yml
+++ b/.github/actions/install-tackle/action.yml
@@ -1,0 +1,49 @@
+name: Install tackle operator
+description: |
+  Install Tackle Operator.
+inputs: 
+  tackle-operator-index-image:
+    description: "image url for operator index"
+    required: false
+    default: "quay.io/konveyor/tackle2-operator-index:latest"
+  tackle-hub-image:
+    description: "image url for tackle-hub"
+    required: false
+    default: "quay.io/konveyor/tackle2-hub:latest"
+  tackle-pathfinder-image:
+    description: "image url for pathfinder image"
+    required: false
+    default: "quay.io/konveyor/tackle-pathfinder:1.3.0-native"
+  tackle-ui-image:
+    description: "image url for tackle-ui"
+    required: false
+    default: "quay.io/konveyor/tackle2-ui:latest"
+  tackle-addon-admin-image:
+    description: "image urd for admin addon"
+    required: false
+    default: "quay.io/konveyor/tackle2-addon:latest"
+  tackle-addon-windup-image:
+    description: "image url for windup addon"
+    required: false
+    default: "quay.io/konveyor/tackle2-addon-windup:latest"
+  # TODO(djzager): uncomment when supported
+  # tackle-operator-container-image:
+  #   description: "image url for operator container"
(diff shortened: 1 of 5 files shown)


.github/actions/install-tackle/action.yml
```diff
@@ -0,0 +1,49 @@
+name: Install tackle operator
```
djzager, line 1: I do wonder if this should be two actions, one for installing the operator and another for creating the tackle CR.

djzager:
I figure that if I can make these scripts work on my extremely slow network. It should be ready for primetime :sunglasses:

fbladilo:
@djzager They look good assuming they've gotten a good round of testing , I have these questions :

Makefile
```diff
@@ -192,3 +192,11 @@ catalog-build: opm ## Build a catalog image.
 .PHONY: catalog-push
 catalog-push: ## Push a catalog image.
 	$(MAKE) docker-push IMG=$(CATALOG_IMG)
+
+.PHONY: start-minikube
+start-minikube:
```
fbladilo, line 197: @djzager We gotta ensure these get reconciled when it is time to re-scaffold the operator

djzager: Speaking in ignorance, is that something to do with a future changes required when picking up newer ansible-operator versions?

fbladilo: When we pickup newer ansible-operator , we tend to re-scaffold in that newer version which causes Makefile among many other files to be overwritten , so that's why is a good idea at least to make a mental note of the new targets for CI.

.github/actions/install-tackle/action.yml
djzager: I think the right approach, in a future PR, would be to modify the CSV directly similar to how they do it in [OpenShift CI](https://docs.ci.openshift.org/docs/how-tos/testing-operator-sdk-operators/#building-operator-bundles) then rebuild the bundle + index and run that in minikube. That would give us the ability to test `tackle2-operator` with the same action.

fbladilo (approved):

PR #115, opened 2022-09-28
Upgrade keycloak to v18

fbladilo:
- Align upstream keycloak release with RHSSO 7.6 (based on v18)
- Targeted for 2.1.2

References in RHSSO 7.6 announcement : 

[RHSSO 7.6](https://access.redhat.com/announcements/6965643)

### DIFF
--- a/bundle/manifests/tackle-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/tackle-operator.clusterserviceversion.yaml
@@ -11,7 +11,7 @@ metadata:
     support: https://github.com/konveyor/tackle2-operator/issues
     repository: https://github.com/konveyor/tackle2-operator
     certified: "false"
-    containerImage: quay.io/fbladilo/tackle2-operator:latest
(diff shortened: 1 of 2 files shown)


jortel (approved):
LGTM

mturley:
@fbladilo thanks for the heads up. In the UI we are using the `keycloak-js` client npm package published by keycloak, so hopefully all we'll need to do is upgrade that dependency to match. I think the actual `redirect_uri` param is abstracted away.

fbladilo:
> @fbladilo thanks for the heads up. In the UI we are using the `keycloak-js` client npm package published by keycloak, so hopefully all we'll need to do is upgrade that dependency to match. I think the actual `redirect_uri` param is abstracted away.

No problem, hopefully is straightforward , it is easy to reproduce by just attempting to log off a user and should see the error.

gildub:
@fbladilo, @mturley, the redirect_uri is not used by tackle2-ui and neither by tackle 1 at least when we took over.

PR #116, opened 2022-09-30
Expose APP_NAME to HUB and UI

fbladilo:
- APP_NAME defines the naming prefix to all user facing objects (pods, secrets, etc)
- APP_NAME is declared within the bundle CSV with a default of "tackle" , if the variable is not defined in CSV it also defaults to "tackle"
- APP_NAME exposed to UI and HUB deployments, they can ingest this environment variable value when creating objects or any other related purposes


jortel (approved):
LGTM

PR #122, opened 2022-10-07
Ensure PG pods readiness on deployment and other fixes

fbladilo:
- Deprecate community.kubernetes collection in favor of kubernetes.core in operator, this is needed in order to use latest functionality in k8s and k8s_info modules, use fqmn in tasks where needed.
- Prevents keycloak and pathfinder restarts due to races with PG database initialization on first deployment
- Always set pathfinder deployment with QUARKUS noauth profile
- Remove creation of tackle admin user , the HUB will take care of this
- Annotate service serving certs only when TLS is enabled
- Define consistent URLs for all components
- Reconcile Developer perspective topologies
- Add a valid spec to Tackles in ALM examples
- Re-sync CSV templates
- Needs backport v2.1.2


mrnold (approved):

rayfordj (approved):
/lgtm

PR #143, opened 2023-01-12
Closes #140, Make RWX Volume Optional

### LINKED ISSUE #140: Optional RWX
Integration with: https://github.com/konveyor/tackle2-hub/issues/148

Goals:
- Only create CACHE PVC when supported by the cluster.
- Discontinue support for Addon CR to list volumes to be mounted.
- Support concept of a _general_ CACHE (RWX) volume that is mounted on the Hub (by the operator) and each task pod (by the task manager).

Changes:
- Add new operator setting: `rwx_supported` (bool).  Default: true
- Rename all `maven-` settings to `rwx-`.
- Rename all `tackle-maven-` resources to `tackle-cache-`.
- The maven PVC no longer injected into the windup addon CR.
- When RWX is supported:
  - Create PV and PVC `tackle-cache-` and mount in the Hub.
  - Set RWX_SUPPORTED Envar in the Hub and UI.
  - Set CACHE_PVC in the Hub to the name of the PVC.
  - Mount the bucket volume RWO in the Hub only when it does not exist.
 - When RWX not supported:
   - Mount the bucket volume RWO in the Hub.  **Existing PVC must not be updated to preserve data in upgrades.
- The addon CRD needs to be updated.
- The _admin_ addon no longer installed. CR to be removed.

jmontleon:


roles/tackle/tasks/main.yml
```diff
@@ -286,24 +300,27 @@
     definition: "{{ lookup('template', 'route-ui.yml.j2') }}"
   when: openshift_cluster|bool
 
-- name: "Setup Maven PersistentVolumeClaim"
+- name: "Setup RWX PersistentVolumeClaim"
```
jortel, line 303: s/RWX/Cache

roles/tackle/templates/deployment-hub.yml.j2
```diff
@@ -148,6 +148,14 @@ spec:
 {% if hub_task_retries is defined %}
             - name: TASK_RETRIES
               value: "{{ hub_task_retries }}"
+{% endif %}
+            - name: RWX_SUPPORTED
+              value: "{{ rwx_supported | string | lower }}"
+{% if rwx_supported %}
+            - name: RWX_PVC
```
jortel, line 155: s/RWX/CACHE

roles/tackle/templates/deployment-hub.yml.j2
```diff
@@ -148,6 +148,14 @@ spec:
...
+              value: "{{ cache_data_volume_claim_name }}"
+            - name: RWX_PATH
```
jortel, line 157: s/RWX/CACHE

jmontleon:
Thanks for catching those. They should be fixed now.

jortel:
I think we still need to copy the updated Addon CRD.

jortel:
This works well for me.

jortel (approved):
LGTM

jortel:
@fbladilo Can you please review?  Thanks!

roles/tackle/templates/persistentvolumeclaim-hub-bucket.yml.j2
```diff
@@ -11,13 +11,13 @@ metadata:
...
   accessModes:
-    - ReadWriteMany
+    - {{ hub_bucket_volume_claim_mode }}
   resources:
     requests:
       storage: {{ hub_bucket_volume_size }}
-{% if rwx_storage_class is defined %}
-{% if rwx_storage_class|length %}
-  storageClassName: {{ rwx_storage_class }}
+{% if cache_storage_class is defined %}
```
fbladilo, line 18: @jmontleon So use for rwx_storage_class used to encompass the same access mode for maven and bucket , now that this is not the case. I think bucket_storage_class should have its own controlling variable instead of using cache_storage_class.

Existing customers using rwx_storage_class set might see their requested storage class name ignored from now on

roles/tackle/defaults/main.yml
```diff
@@ -167,10 +168,13 @@ windup_service_name: "{{ app_name }}-{{ windup_name }}-{{ windup_component_name
...
 
-maven_name: "maven"
-maven_data_volume_size: "100Gi"
-maven_data_volume_name: "{{ app_name }}-{{ maven_name }}-data"
-maven_data_volume_claim_name: "{{ app_name }}-{{ maven_name }}-volume-claim"
+cache_name: "cache"
+cache_data_volume_size: "100Gi"
+cache_data_volume_name: "{{ app_name }}-{{ cache_name }}-data"
+cache_data_volume_claim_name: "{{ app_name }}-{{ cache_name }}-volume-claim"
+cache_data_volume_mode: "ReadWriteMany"
```
fbladilo, line 175: perhaps cache_data_volume_claim_mode to make it consistent with rest of components

roles/tackle/templates/persistentvolumeclaim-hub-bucket.yml.j2
jmontleon: My understanding is that there is nothing to be gained by having the hub bucket volume created as RWX, which is why it's set to RWO by default now.

As mentioned above I'm aware the accessMode is immutable and there may be existing installs with it set as RWX. That's why we check and override the value with the existing value. New installs should be created as RWO.

Is there a reason we expect a user should want or need to create the hub bucket as an RWX volume? I guess technically they could override it with `hub_bucket_volume_claim_mode`, though my motivation to parameterize this  was more for the sake of allowing the playbook to complete without error when and pv with the old access mode already exists.

roles/tackle/defaults/main.yml
jmontleon: sure!

roles/tackle/templates/persistentvolumeclaim-hub-bucket.yml.j2
jmontleon: @fbladilo I think I misunderstood what you were saying. I see we have the cache params in the hub bucket PV. Let me fix.

jmontleon: @fbladilo I changed it to hub_bucket_storage_class, which I think is consistent with the other hub bucket parameter names. Let me know if you prefer something else or see any others I missed. Thanks!

jmontleon: @fbladilo I wonder if we should do something about your comment regarding rwx_storage_class.... we could do something like if rwx_storage_class defined set_fact cache_storage_class... maybe check that cache_storage_class is not defined in case they become wise to the change and somehow end up with both set... let me know your thoughts, I don't have a strong opinion, so happy to implement whatever you think is best.

roles/tackle/templates/persistentvolumeclaim-hub-bucket.yml.j2
```diff
@@ -11,13 +11,13 @@ metadata:
...
+    - {{ hub_bucket_volume_claim_mode }}
   resources:
     requests:
       storage: {{ hub_bucket_volume_size }}
-{% if rwx_storage_class is defined %}
-{% if rwx_storage_class|length %}
-  storageClassName: {{ rwx_storage_class }}
+{% if hub_bucket_storage_class is defined %}
+{% if hub_bucket_storage_class|length %}
+  storageClassName: {{ cache_storage_class }}
```
fbladilo, line 20: should be set to hub_bucket_storage_class

fbladilo:
@jmontleon We should also probably update README with the new variables regarding RWX

https://github.com/konveyor/tackle2-operator#tackle-cr-settings

fbladilo (approved):
LGTM , thanks @jmontleon !

PR #146, opened 2023-01-23
Use maven_data_volume_size when cache_data_volume_size is not set

jmontleon:


fbladilo (approved):
LGTM , rwx_storage_class is already being transferred , so we provide legacy coverage there.

PR #149, opened 2023-01-25
Windup addon ensure image pull policy is set

fbladilo:
The windup addon image pull policy is now controlled by the global image_pull_policy parameter in the Tackle CR. 

This addresses possible inconsistencies between the windup addon pull policy and the rest of Tackle components.

### DIFF
--- a/roles/tackle/templates/customresource-addon-windup.yml.j2
+++ b/roles/tackle/templates/customresource-addon-windup.yml.j2
@@ -10,6 +10,7 @@ metadata:
     app.kubernetes.io/part-of: {{ app_name }}
 spec:
   image: {{ windup_fqin }}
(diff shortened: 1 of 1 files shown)


jortel (approved):
LGTM

jortel:
@djzager @jmontleon IIRC the addon CI test needs an IfNotPresent policy because it builds an image in the local docker registry.  The default for the `image_pull_policy` in the operator is `Always`. The Tackle CR used in the test will need to be updated to override to `image_pull_policy: IfNotPresent`.

djzager:
The windup workflow is already using it https://github.com/konveyor/tackle2-addon-windup/blob/50a60382a4ab39ae86e5247680df277f6d9e14af/.github/workflows/test-windup.yml#L51 thanks for the reminder @jortel

PR #152, opened 2023-02-10
Ensure windup addon is always reconciled

fbladilo:

The windup addon (or any other addon for this matter), should always be reconciled to avoid stale data , such as a old addon image reference or to also honor Tackle CR settings such as image pull policy or resource requests.

This is specially impactful during Tackle upgrades, see : https://issues.redhat.com/browse/MTA-127

This PR removes the set and forget task by operator and forces to reconcile in every cycle.


jortel (approved):
LGTM

PR #154, opened 2023-02-13
:sparkles: Build operator+bundle using Composite GH Action

jmontleon:
Signed-off-by: Jason Montleon <jmontleo@redhat.com>

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,41 @@
+name: 'Publish'
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - 'main'
+      - 'release-*'
+    tags:
+      - 'v*'
+
+jobs:
+  push-quay:
+    name: Build and Push Manifest
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+    steps:
+    - name: Build Operator Container
+      uses: konveyor/release-tools/build-push-quay@main
+      with:
+        architectures: "amd64, arm64, ppc64le, s390x"
+        containerfile: "./Dockerfile"
+        image_name: "tackle2-operator"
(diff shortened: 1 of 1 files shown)


.github/workflows/bundle-image-build-push.yml
```diff
@@ -0,0 +1,29 @@
+name: 'Build and Push Bundle Image'
```
djzager, line 1: I think we can reduce the total number of workflows we have in the end if we were to name this one `.github/workflows/publish.yml`, wdyt?

The workflow.name is important, maybe "Publish Operator" (wouldn't be surprised if after making a release this workflow runs to build the images and push to community operators | helm charts.

.github/workflows/bundle-image-build-push.yml
```diff
@@ -0,0 +1,29 @@
...
+      - 'v*'
+
+jobs:
+  push-quay:
+    name: Build and Push Manifest
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+    steps:
+    - name: Checkout Push to Registry action
```
djzager, line 19: Wouldn't we want the bundle step to run after the operator step to make sure, when we make a `v2.1.0-alpha.1`, that the container image exists before we bundle it?

jmontleon: I don't think it will matter much for these two. My feeling is we need all the artifacts to be ready before we generate or update an index image and/or create a PR for community operators. But as far as I know there's nothing as far as the operator container and bundle images that are dependent on the others.

@shawn-hurley do you have an opinion?

jmontleon: Actually I take it back. We will need the SHA, so yes, after.

djzager (approved):

PR #157, opened 2023-02-21
:bug: Manage DB container for RHSSO

jmontleon:
These changes are intended to switch us from using a postgresql container provisioned by the RHSSO operator which is unsupported and uses a tag making it impossible to use MTA in a disconnected cluster (without enabling an experimental feature to allow mirroring with tags).

The operator will cleanly switch the operator to use a postgresql database container provisioned by us, but it will not migrate data. Users will be required to log back with the default password and reconfigure any users they have created.

~~Up to this point as far as I can tell RHSSO does not appear to deprovision its postgresql deployment when the configuration is changed to use an 'external' database and the keycloak-discovery service never becomes ready, so for an upgrade these manual steps are currently required. I'm still looking to see if we can smooth this out and have some ideas. For example, we can probably look at the mta-rhsso keycloak resource status for the keycloak-postgresql deployment and if it exists nuke it and recreate it, which should clean up keycloak and reprovision it properly.~~
  
~~oc delete keycloak -n openshift-mta mta-rhsso --ignore-not-found=true && \
oc scale deployment -n openshift-mta mta-operator --replicas=0 && \
oc scale deployment -n openshift-mta mta-operator --replicas=1~~

### DIFF
--- a/roles/tackle/tasks/main.yml
+++ b/roles/tackle/tasks/main.yml
@@ -42,7 +42,6 @@
 
 - when:
     - feature_auth_required|bool
-    - app_profile == "konveyor"
   block:
     - name: "Setup Keycloak PostgreSQL PersistentVolumeClaim"
       k8s:
@@ -91,7 +90,7 @@
         namespace: "{{ app_namespace }}"
         label_selectors:
           - app.kubernetes.io/name = {{ keycloak_database_service_name }}
-        wait: yes
+        wait: true
         wait_condition:
           type: "Ready"
           status: "True"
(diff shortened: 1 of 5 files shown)


roles/tackle/tasks/main.yml
```diff
@@ -120,33 +119,87 @@
...
+      k8s_info:
+        api_version: v1
+        kind: Secret
+        name: "{{ keycloak_database_secret_name }}"
+        namespace: "{{ app_namespace }}"
+      register: keycloak_database_secret
+
+    - name: "Get DNS operator CR"
+      k8s_info:
+        api_version: operator.openshift.io/v1
```
jwmatthews, line 169: How would this work with non-OpenShift: Keycloak on a vanilla k8s env?

Assume we need to at least wrap this with something to check for presence of this operator?

jwmatthews: I see now the logic is that "mta" differentiates what is invoked so "app_profile == "mta" block is for just the OpenShift case

roles/tackle/templates/secret-keycloak-db.yml.j2
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
+data:
+  POSTGRES_DATABASE: {{ rhsso_db_name_b64 }} 
+  POSTGRES_EXTERNAL_ADDRESS: {{ rhsso_db_host_b64 }}
+  POSTGRES_HOST:
```
fbladilo, line 5: Should we get rid of POSTGRES_HOST here, seems like for external only POSTGRES_EXTERNAL_ADDRESS is used correct?

roles/tackle/templates/secret-keycloak-db.yml.j2
```diff
@@ -0,0 +1,19 @@
...
+  POSTGRES_PASSWORD: {{ rhsso_db_pass_b64 }}
+  POSTGRES_USERNAME: {{ rhsso_db_user_b64 }}
+  POSTGRES_SUPERUSER: {{ "true" | b64encode }}
+  SSLMODE: {{ "prefer" | b64encode }}
```
fbladilo, line 9: Does prefer mean we attempt TLS and fall back to non-TLS on failure?

roles/tackle/templates/deployment-hub.yml.j2
```diff
@@ -39,6 +39,11 @@ spec:
         app.kubernetes.io/part-of: {{ app_name }}
         app: {{ app_name }}
         role: {{ hub_service_name }}
+{% if feature_auth_required|bool %}
+{% if app_profile == 'mta' %}
+        keycloak_db_secret_name: {{ keycloak_db_secret.env | k8s_config_resource_name }}
```
fbladilo, line 44: Curious what would this label be used for?

roles/tackle/templates/secret-keycloak-db.yml.j2
jmontleon: No. It is intentional and blanks the value inserted by the RHSSO operator when it uses its own DB.

roles/tackle/templates/secret-keycloak-db.yml.j2
jmontleon: Yes. I believe from the postgres documentation this is the default. From what I could see what it was doing running its own DB is using `""` as the value so the same.

This lands in this env var on the keycloak DB and mu understanding is that it is a client side setting:
```
    - name: DB_CONNECTION_PROPERTY_sslmode
      value: prefer
```

`require` has the same meaning as `verify-ca` when a CA is  passed in, which it seems they do and therefore does not work.

This is honestly not great, but not worse than what we have today, I don't believe upstream or down. If that's true I'd rather we file an issue to improve the pgsql deployment upstream and down. If there's a quick win to fixing it here I'm open to that, but also still filing the issue for upstream keycloak and pathfinder.

jmontleon: @jwmatthews @dymurray do you have a strong feeling about how/when we handle this? 

Depending how the 'client' to pathfinder is configured it may be that there is no SSL, SSL is required and not verifying, or is doing it all already.

roles/tackle/templates/deployment-hub.yml.j2
jmontleon: There has always been the issue that you update an underlying config for a deployment, be that a secret, configmap, the entire authentication system, that we won't be picked up until the pod restarted. The way we handled this for mig-controller in MTC when we found them was to add these. This generates a label:
```
keycloak_db_secret_name: keycloak-db-secret-c74gh8bdd5
```
The last bit changes based on the contents of the calculated contents of the file. We use the template so status, etc. doesn't affect it. Anytime the operator would update the file, the generated name changes, the label is updated, the pod restarts.

I believe fabianvf taught us this. You can see more examples:
https://github.com/migtools/mig-operator/blob/master/roles/migrationcontroller/templates/controller.yml.j2#L45-L47

roles/tackle/templates/secret-keycloak-db.yml.j2
jmontleon: Docs are here: https://www.postgresql.org/docs/current/libpq-ssl.html

roles/tackle/templates/secret-keycloak-db.yml.j2
fbladilo: :heavy_check_mark:

roles/tackle/templates/secret-keycloak-db.yml.j2
fbladilo: Cool , as long as fallback is guaranteed then we should be ok, I believe pathfinder on the client side is probably going unencrypted by default , or perhaps falls back as well.

fbladilo (approved):

PR #159, opened 2023-02-22
:sparkles: Provide openshift-oauth as an authentication option

jmontleon:
WIP. Need to test downstream and turning auth on and off and switching a little more.

Right now I've switched from keycloak -> oauth without issue.

Could use clean up around conditionals in templates as well. These should be possible to make smaller and cleaner.

### DIFF
--- a/roles/tackle/tasks/main.yml
+++ b/roles/tackle/tasks/main.yml
@@ -42,6 +42,56 @@
 
 - when:
     - feature_auth_required|bool
+    - feature_auth_type == "oauth"
+  block:
+    - name: "Check if Cookie Secret already exists"
+      k8s_info:
+        api_version: v1
+        kind: Secret
+        name: cookie-secret
+        namespace: "{{ app_namespace }}"
+      register: cookie_secret
+
+    - name: "Generate Cookie Secret"
+      set_fact:
+        new_cookie_secret: "{{ lookup('password', '/dev/null chars=ascii_lowercase,ascii_uppercase,digits length=32') }}"
+      when: (cookie_secret.resources | length) == 0
+
+    - name: "Create Cookie Secret"
+      k8s:
+        state: present
+        definition: "{{ lookup('template', 'secret-cookie-secret.yml.j2') }}"
+      when: (cookie_secret.resources | length) == 0
+
+    - name: "Retrieve Cookie Secret"
+      k8s_info:
+        api_version: v1
+        kind: Secret
+        name: cookie-secret
+        namespace: "{{ app_namespace }}"
+      register: cookie_secret
+
+    - name: "Set Cookie Secret"
+      set_fact:
+        cookie_secret_data: "{{ cookie_secret.resources[0].data['cookie-secret'] | b64decode }}"
+
+    - name: "Retrieve Oauth Client Secret if it exists"
+      k8s_info:
+        api_version: v1
+        kind: Secret
+        name: oauth-client-secret
+        namespace: "{{ app_namespace }}"
+      register: oauth_client_secret_status
+
+    - name: "Set Oauth Client Secret"
+      set_fact:
+        oauth_client_secret: "{{ oauth_client_secret_status.resources[0].data['client-secret'] | b64decode }}"
+      when: (oauth_client_secret_status.resources | length) > 0
+
+- when:
+    - feature_auth_required|bool
+    - feature_auth_type == "keycloak"
+    - app_profile == "konveyor"
   block:
     - name: "Setup Keycloak PostgreSQL PersistentVolumeClaim"
       k8s:
@@ -134,6 +184,7 @@
 
 - when:
     - feature_auth_required|bool
+    - feature_auth_type == "keycloak"
     - app_profile == "mta"
   block:
     - name: "Check for existing RHSSO Keycloak CR"
@@ -356,6 +407,11 @@
     definition: "{{ lookup('template', 'deployment-hub.yml.j2') }}"
     merge_type: merge
 
+- name: "Setup UI ServiceAccount"
+  k8s:
+    state: present
+    definition: "{{ lookup('template', 'serviceaccount-ui.yml.j2') }}"
+
 - name: "Setup UI Service"
   k8s:
     state: present
@@ -365,6 +421,7 @@
   k8s:
     state: present
     definition: "{{ lookup('template', 'deployment-ui.yml.j2') }}"
+    merge_type: merge
 
 - name: "Setup UI Ingress"
   k8s:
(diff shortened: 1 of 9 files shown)


roles/tackle/templates/deployment-ui.yml.j2
```diff
@@ -37,6 +37,32 @@ spec:
...
+{% if feature_auth_required|bool %}
+{% if feature_auth_type == "openshift-oauth" %}
+        - args:
+          - --https-address=:8443
+          - --provider=openshift
+          - --tls-cert=/etc/tls/private/tls.crt
+          - --tls-key=/etc/tls/private/tls.key
+          - --upstream=http://localhost:8080
+          - --openshift-service-account={{ ui_serviceaccount_name }}
+          - --openshift-sar={"namespace":"{{ app_namespace }}","resource":"services","resourceName":"{{ ui_service_name }}","verb":"get"}
```
shawn-hurley, line 49: I wonder if there is a better approximation? but honestly don't know

jmontleon: Is this in reference to the openshift-sar rule? I took this from the example at https://github.com/openshift/oauth-proxy#openshift-oauth-proxy

If there's something better I am happy to change it.

roles/tackle/templates/serviceaccount-ui.yml.j2
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ ui_serviceaccount_name }}
+  namespace: {{ app_namespace }}
+  annotations:
+    serviceaccounts.openshift.io/oauth-redirectreference.primary: '{"kind":"OAuthRedirectReference","apiVersion":"v1","reference":{"kind":"Route","name":"{{ ui_route_name }}"}}'
```
jmontleon, line 7: Normally I would argue OLM should create an SA, but it needs this annotation for the oauth-proxy to work, so the template pretty much needs to exist in full, even to add the annotation...

roles/tackle/templates/deployment-ui.yml.j2
jmontleon: At the least we can probably make this or an alternative the value for a default `oauth_openshift_sar`. If someone then wants to get inventive and change the sar rule for their environment they can do so.

shawn-hurley: that makes sense, I was just thinking about what it would be that the users care to give access to to allow someone to access the UI. I imagine that this will be fine for the default.

roles/tackle/templates/serviceaccount-ui.yml.j2
fbladilo: Perhaps that's the way there , we let OLM manage the service account itself for the sake of consistency (and cleanliness) , then patch it when openshift-oauth is needed/selected by user ?

jmontleon: I have no problem with that. Right now it'll just be created by the template as tackle-ui, mta-ui or whatever based on profile. It'd probably have to be updated to mta-ui for downstream in the CSV? If that isn't a problem I'm OK with adding it to the CSV. As far as I know it requires no special permissions so it'd be one line.

jmontleon:
I started cleaning up the conditionals. I also made the oauth option more generic. The openshift oauth-proxy image only supports openshift, but it's possible to use a different image and use other providers.

I've had success with OpenShift, Gitea, and GitHub.

Example OpenShift config:
```
spec:
  feature_auth_required: "true"
  feature_auth_type: oauth
```

Example GitHub Config
```
spec:
  feature_auth_required: "true"
  feature_auth_type: oauth
  oauth_access_rule: --github-user="jmontleon"
  oauth_client_id: REDACTED
  oauth_email_domain: '*'
  oauth_image_fqin: quay.io/jmontleon/oauth2-proxy:latest
  oauth_provider: github
```

Example Gitea Config:
```
spec:
  feature_auth_required: "true"
  feature_auth_type: oauth
  oauth_client_id: REDACTED
  oauth_email_domain: '*'
  oauth_extra_opts:
  - --redirect-url=https://tackle-konveyor-tackle.apps.openshift4.example.com/oauth2/callback
  - --provider-display-name="Gitea"
  - --login-url=https://gitea.example.com/login/oauth/authorize
  - --redeem-url=https://gitea.example.com/login/oauth/access_token
  - --validate-url=https://gitea.example.com/api/v1
  - --ssl-upstream-insecure-skip-verify=true
  - --ssl-insecure-skip-verify=true
  oauth_image_fqin: quay.io/oauth2-proxy/oauth2-proxy:v7.2.0
  oauth_provider: github
```

You can store the oauth client secret in a secret rather than placing it in the CR:
```
apiVersion: v1
data:
  client-secret: b64-encoded-secret
kind: Secret
metadata:
  name: oauth-client-secret
  namespace: konveyor-tackle
type: Opaque
```

There are unfortunately bugs with both of these alternate providers, requiring using newer/older images.
https://github.com/oauth2-proxy/oauth2-proxy/issues/1724
https://github.com/oauth2-proxy/oauth2-proxy/pull/1996

I'll work on more cleanup.

shawn-hurley:
Is there something as a followup to enable this to also be how we do authorization? @jortel @mansam IIUC we have a specific auth tied to keycloack in the hub API? 

I am also wondering if there is any authorization that the UI does? @ibolton336? 

I wonder if we can track this work in issues as follow ons and prioitze them accordingly?

ibolton336:
> Is there something as a followup to enable this to also be how we do authorization? @jortel @mansam IIUC we have a specific auth tied to keycloack in the hub API?
> 
> I am also wondering if there is any authorization that the UI does? @ibolton336?
> 
> I wonder if we can track this work in issues as follow ons and prioitze them accordingly?

@shawn-hurley The UI authentication is using [keycloak-js](https://www.keycloak.org/docs/latest/securing_apps/index.html#_javascript_adapter) and relies almost entirely on the keycloak instance being created elsewhere. We are just interfacing with it via this keycloak adapter.

jmontleon:
@ibolton336 as far as the UI is concerned this is pretty much like a noauth scenario.

See: https://www.youtube.com/watch?v=DKmdYMwePp0&feature=youtu.be

There's no easy way to log out now. It might be an good for an enhancement if we have UI enable to more easily facilitate logout. I think it would basically be directing to `/oauth2/sign_out`, but would have to confirm.

jmontleon:
I've marked it ready for review. I've used it on a few occasions and it works fine for me. Would be nice if one or two others can confirm working results.

roles/tackle/templates/deployment-ui.yml.j2
```diff
@@ -37,6 +35,51 @@ spec:
         role: {{ ui_service_name }}
     spec:
       containers:
+{% if feature_auth_required|bool and feature_auth_type == "oauth" %}
```
fbladilo, line 38: Agreed the conditionals get crazy in the UI deployment , I'm wondering if perhaps there is away in jinja2 to pre-assemble or pre-process based on selected settings somewhere else and just dump the final contents in this deployment

jmontleon: @fbladilo I was complaining earlier on in the thread about nested conditionals before I did cleanup in some of my prior rebases. Taking this for example:

```
{% if feature_auth_required|bool %}
{% if feature_auth_type == "oauth" %}
...
{% endif %}
{% endif %}
```

In some places they were three and four deep and very hard to follow.

If you see an obvious way to clean up further I'm happy to do so.

roles/tackle/tasks/main.yml
```diff
@@ -410,3 +467,30 @@
     state: present
     definition: "{{ lookup('template', 'networkpolicy.yml.j2') }}"
   when: feature_isolate_namespace|bool
+
+- when:
+    - not(feature_auth_required|bool) or not(feature_auth_type == "keycloak")
+  block:
+    - name: "Deprovision RHSSO Keycloak CR"
```
shawn-hurley, line 474: Does this by itself not remove everything?

roles/tackle/templates/deployment-ui.yml.j2
shawn-hurley: I think having whole files that Ansible chooses can sometimes be nice if the conditionals are super nested and hard to follow?

shawn-hurley (approved):
I think this looks good to me visually, the only thing that I think we may want to talk about, is do we want to give * on service accounts for permissions, maybe just create/update?

IDK just a thought

roles/tackle/tasks/main.yml
jmontleon: This block does. It removes the keycloak configuration, the SSO deployment, and the DB deployment. I left the PVC alone so we don't destroy data. If someone wanted to try Oauth or switch to auth false for some reason and then decides to switch back their configured users will be as they left them, etc.

PR #161, opened 2023-03-08
:seedling: Action to automate publishing Konveyor operator releases

### LINKED ISSUE #155: Workflow to publish to community-operators
We need a workflow, executed when needed (ie. workflow_dispatch), that would submits a PR to https://github.com/k8s-operatorhub/community-operators/tree/main/operators . The constraints:

* We want sha digest pinning. This means that each of the relatedImages must be mirrored to quay with a particular tag (to keep the sha from being garbage collected). Image tags in the CSV should be updated to use sha digests
* Should accept the version (ie. `vX.Y.Z-abcd.1234`) as input

jmontleon:
Fixes #155

### DIFF
--- a/.github/workflows/march-image-build-push.yml
+++ b/.github/workflows/march-image-build-push.yml
@@ -1,0 +1,45 @@
+name: 'Build and Push Multi-Arch Image'
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - 'main'
+      - 'release-*'
+    tags:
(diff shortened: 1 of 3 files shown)


jmontleon:
Running the dispatch action to post a PR for a new release would look like so (plus filling in the previous version when there is one)

image

djzager (approved):

djzager (approved):
LGTM

PR #162, opened 2023-03-13
:bug: Fix keycloak deployment check

jmontleon:
When not coming from an upgrade situation the Deployment array in secondary resources doesn't exist and it causes a reconcile failure.


rayfordj (approved):
/lgtm

PR #166, opened 2023-03-15
:bug: Update memory limit for minikube start command in the documentation

savitharaghunathan:
Following the install guide as-is throws an error while starting minikube. To solve the error, this PR adjusts the memory limit for minikube to be a little less than podman.


fbladilo (approved):
ACK

PR #173, opened 2023-03-29
:bug: Set fsGroup for non-OpenShift clusters

jmontleon:
On EKS and probably other kubernetes variants the PostgreSQL pods run as user 26. Without setting the fsGroup the storage is mounted 755 owned by root:root so postgres can't write and fails to start with a permission denied.

26 comes from the Dockerfile the image is built with https://github.com/sclorg/postgresql-container/blob/master/12/Dockerfile.centos7#L88


fbladilo (approved):
LGTM , @jmontleon In OpenShift I assume the USER 26 setting gets ignored in favor of the ephemeral assigned user correct?

jmontleon:
@fbladilo correct.

PR #175, opened 2023-03-30
:seedling: create-release workflow should pass token

djzager:
We can't use GitHub's automatic token if 1) we want to create releases in other repositories (ie. operator creating a release in the hub) and 2) we want the publish steps to run in those repos.


jmontleon (approved):
/LGTM

PR #176, opened 2023-04-02
:bug: Adjust Windup Addon Requests and Add Limits

jmontleon:
Our requests are extremely inaccurate compared to what the container can/will use. From what I've observed each container pegs a CPU at pretty much 100%, and RAM usage far exceeds 2Gi. 

The most immediate thing we can do to help prevent bulk analysis from toppling nodes is to increase the requests for both of these.

Other things we should consider:
- Adding an env var that equals half the requests value the windup pod to limit the JVM RAM usage. My belief that this should be half comes from the description of MAX_MEMORY which reads as if additional NON-HEAP Memory will be allocated equal to the HEAP memory specified by this var.
https://www.ibm.com/docs/en/coss/3.14.0?topic=application-change-memory-settings
```
Maximum memory the daemon should allocate for its heap when configuring memory dynamically.

Note: An equivalent amount will be used for off-heap memory unless configured otherwise....
```
This may effectively act as a limit so the next point may not be necessary for constraining memory utilization.
- Implementing adjustable LIMITS on the windup pods in addition to requests
- Linking to documentation and advising customers to implement eviction thresholds to protect their nodes.
https://docs.openshift.com/container-platform/4.12/nodes/nodes/nodes-nodes-resources-configuring.html#allocate-eviction-thresholds_nodes-nodes-resources-configuring

I opened an issue on this last point because the eviction thresholds did not work as it was suggested they would in the documentation.
https://issues.redhat.com/browse/OCPNODE-1579


shawn-hurley:
I think we need to do this now and it LGTM. 

I am still a plus one on limits being set and configurable.

shawn-hurley:
Plus one on the limits, @jmontleon  did we run in a small Minikube cluster

jmontleon:
@jortel I am not sure if half is the right answer. The bit I pasted above implies to me that it will allocate additional non-HEAP memory equal to the value set here for HEAP, but then digging again I noticed valid values for MAX_MEMORY go up to 32000 while there is a memory value that only goes up to 16000 and states, "Set the heap size to a fixed value. An equal amount of memory is allocated for off-heap memory if not configured otherwise."

https://www.ibm.com/docs/en/coss/3.14.0?topic=application-change-memory-settings

I honestly don't know how the wording for what should be such straight forward options can be written to leave so much question. Maybe you can make some sense of it.

jortel:
@jmontleon I'm not convinced we have an immediate problem to solve.
In light of https://medium.com/marionete/managing-java-heap-size-in-kubernetes-3807159e2438.

PR #178, opened 2023-04-03
:seedling: create-release: Add remaining projects

djzager:
In addition to adding the remainder of the projects we are concerned with. This also includes a "wait" for each of the images to be published. We can later use this as our gate for the publish step.


jmontleon (approved):
ACK

PR #192, opened 2023-04-12
:seedling: Allow overriding to use podman instead of docker in Makefile

jwmatthews:
I ran into issues building on mac and having the Makefile honor the alias docker=podman


jmontleon (approved):
/LGTM

PR #195, opened 2023-04-13
:seedling: Updated for konveyor-operator name change for dev workflow

jwmatthews:
The [tools/tackle-opdev.sh](https://github.com/konveyor/tackle2-operator/blob/main/tools/tackle-opdev.sh#L155) script uses this for dev workflows

I think this needs to updated for the konveyor-operator rename.


jmontleon (approved):
ACK

PR #196, opened 2023-04-13
:seedling: Updates to build on mac arm64 and deploy to amd64

jwmatthews:
With these changes I'm able to locally build on my arm64 Mac and deploy the images onto my EKS cluster (amd64).

Needed to address 2 issues:

1. Allow --arch amd64 to be passed into image builds in Makefile
2. Account for "sed -i" difference on Mac vs Linux


jmontleon (approved):
ACK

Issue #170, opened 2023-03-27
Document how to install and access UI from EKS


jwmatthews:
Let's walk through an install of Konveyor to an EKS cluster and document the steps and any tweaks needed for getting Ingress configured.

In particular, let's ensure we cover both flows of auth enabled/disabled, so we can be sure the workflow is accessible when keycloack is enabled.

jwmatthews:
We now have the means of easily spinning up AWS EKS clusters with Ansible + CLIs for (eksctl + aws) to aid other contributors for development/testing.  You will need to supply your own AWS Credentials.

https://github.com/konveyor/hack_env_helpers/tree/main/aws/eks

The above will provision an EKS cluster and configure both the EBSCI AddOn and the ALB Ingress controller.  Roughly ~20 mins to setup cluster and ~5-10 minutes to install software to test it out, including OLM + Konveyor.

This gives us the needed starting point so we have an EKS cluster with dynamic PV provisioning (RWO only, no RWX) and Ingress via AWS LoadBalancer.

First thing noticed is issue #184 where the Ingress we create has a [hard assumption of nginx for Ingress.](https://github.com/konveyor/tackle2-operator/blob/main/roles/tackle/templates/ingress-ui.yml.j2#L15)
```
spec:
  ingressClassName: nginx
```

djzager:
@jwmatthews @jmontleon Looks like PRs have merged that maybe fixed this issue. Where does it stand?

dymurray:
Closing due to inactivity or age, please reopen if you feel this was done in error

PR #197, opened 2023-04-14
:sparkles: Support for Amazon Load Balancer Ingress controller

### LINKED ISSUE #167: Accessing the tackle operator in EKS
I have installed the latest version of tackle but not some pods are in a pending state.

`NAMESPACE            NAME                                                              READY   STATUS      RESTARTS   AGE
default              nginx-deployment-7fb96c846b-cvf4t                                 1/1     Running     0          5h42m
default              nginx-deployment-7fb96c846b-nhp6d                                 1/1     Running     0          5h42m
default              nginx-deployment-7fb96c846b-wbx9h                                 1/1     Running     0          5h42m
konveyor-tackle      konveyor-tackle-wr9w2                                             1/1     Running     0          5h28m
kube-system          aws-node-9nqp7                                                    1/1     Running     0          6h51m
kube-system          aws-node-tsmq4                                                    1/1     Running     0          6h51m
kube-system          cluster-autoscaler-7d58d9969d-75mtf                               1/1     Running     0          4h26m
kube-system          coredns-7cf95b74cc-hdk2z                                          1/1     Running     0          6h57m
kube-system          coredns-7cf95b74cc-x6twf                                          1/1     Running     0          6h57m
kube-system          kube-proxy-t5h5m                                                  1/1     Running     0          6h51m
kube-system          kube-proxy-xx8vj                                                  1/1     Running     0          6h51m
my-tackle-operator   tackle-keycloak-postgresql-f455ffdf8-b5qn9                        0/1     Pending     0          5h19m
my-tackle-operator   tackle-operator-7fc8bd6bb7-87ks7                                  1/1     Running     0          5h18m
olm                  a42881ea1507f1361d60825127fbe483b60b751c575a69ef734fcd24e6ctlfq   0/1     Completed   0          6h20m
olm                  catalog-operator-655fb46cd4-5jnkb                                 1/1     Running     0          6h22m
olm                  olm-operator-67fdb4b99d-txv6c                                     1/1     Running     0          6h22m
olm                  packageserver-5f55f478ff-s2fv4                                    1/1     Running     0          6h21m
muralithimmegowda@CB-L015 ~ % 

tackle-keycloak-postgresql : It has no events or errors but is in a pending state. 

It says succeeded. I have installed this in the latest version of EKS 1.25. 
muralithimmegowda@CB-L015 ~ % kubectl get csv -n my-tackle-operator
NAME                     DISPLAY           VERSION   REPLACES                 PHASE
tackle-operator.v2.1.3   Tackle Operator   2.1.3     tackle-operator.v2.1.2   Succeeded
muralithimmegowda@CB-L015 ~ % 

It has not created any ingress how to access the application. Please suggest.

### LINKED ISSUE #184: EKS with ALB:  Changes required to work with Amazon Load Balancer ingress controller
Related #167 and #170 

On an AWS EKS cluster with the AWS LoadBalancer configured as Ingress, the operator is unable to create the Ingress resource 

[due to
](https://github.com/konveyor/tackle2-operator/blob/main/roles/tackle/templates/ingress-ui.yml.j2#L15)
```
spec:
  ingressClassName: nginx

```

from operator logs
```
--------------------------- Ansible Task StdOut -------------------------------

TASK [tackle : Setup UI Ingress] ***********************************************
task path: /opt/ansible/roles/tackle/tasks/main.yml:369

-------------------------------------------------------------------------------
{"level":"info","ts":1680981644.5812697,"logger":"proxy","msg":"Cache miss: networking.k8s.io/v1, Kind=Ingress, konveyor-tackle/tackle"}
{"level":"info","ts":1680981644.5886374,"logger":"proxy","msg":"Injecting owner reference"}
{"level":"info","ts":1680981644.5889103,"logger":"proxy","msg":"Watching child resource","kind":"networking.k8s.io/v1, Kind=Ingress","enqueue_kind":"tackle.konveyor.io/v1alpha1, Kind=Tackle"}
{"level":"info","ts":1680981644.588938,"msg":"Starting EventSource","controller":"tackle-controller","source":"kind source: *unstructured.Unstructured"}
{"level":"error","ts":1680981644.856265,"logger":"logging_event_handler","msg":"","name":"tackle","namespace":"konveyor-tackle","gvk":"tackle.konveyor.io/v1alpha1, Kind=Tackle","event_type":"runner_on_failed","job":"1812945292459884819","EventData.Task":"Setup UI Ingress","EventData.TaskArgs":"","EventData.FailedTaskPath":"/opt/ansible/roles/tackle/tasks/main.yml:369","error":"[playbook task failed]"}

--------------------------- Ansible Task StdOut -------------------------------

 TASK [Setup UI Ingress] ******************************** 
fatal: [localhost]: FAILED! => {"changed": false, "error": 403, "msg": "Ingress tackle: Failed to create object: b'{\"kind\":\"Status\",\"apiVersion\":\"v1\",\"metadata\":{},\"status\":\"Failure\",\"message\":\"admission webhook \\\\\"vingress.elbv2.k8s.aws\\\\\" denied the request: invalid ingress class: IngressClass.networking.k8s.io \\\\\"nginx\\\\\" not found\",\"reason\":\"invalid ingress class: IngressClass.networking.k8s.io \\\\\"nginx\\\\\" not found\",\"code\":403}\\n'", "reason": "Forbidden", "status": 403}
```

jwmatthews:
Fixes #184 
Fixes #167 (confirmed working with `feature_auth_required: true`)
Relates to #170

These changes allow us to deploy Konveyor onto an Amazon EKS cluster that uses the Amazon recommended Amazon Load Balancer ingress opposed to 'nginx'. 

I deployed the EKS cluster for testing using this newly developed automation:
https://github.com/konveyor/hack_env_helpers/tree/main/aws/eks

I captured notes while exploring ALB Ingress here for debugging purposes:
https://gist.github.com/jwmatthews/d492a4944c83dc3f2d016187d53d13b2

### DIFF
--- a/roles/tackle/defaults/main.yml
+++ b/roles/tackle/defaults/main.yml
@@ -146,11 +146,8 @@ ui_container_limits_cpu: "500m"
 ui_container_limits_memory: "800Mi"
 ui_container_requests_cpu: "100m"
 ui_container_requests_memory: "350Mi"
-# Setting ingress defaults for 'nginx' ingress
-# To use 'alb' ingress need to set:
-#   ui_ingress_class_name: "alb"
-#   ui_ingress_path_type: "Prefix"
 ui_ingress_name: "{{ app_name }}"
+# Ingress class names of "nginx" and "alb" are expected
 ui_ingress_class_name: "nginx"
 ui_ingress_proxy_body_size: "500m"
 ui_route_name: "{{ app_name }}"
@@ -163,7 +160,7 @@ ui_route_tls_termination: "edge"
 ui_route_tls_insecure_termination_policy: "Redirect"
 # ui_ingress_path_type should only be defined if you need to override
 # the default value for the ingress controller you are using
-#ui_ingress_path_type: 
+# ui_ingress_path_type: 
 
 
 oauth_provider: openshift
--- a/roles/tackle/defaults/main.yml
+++ b/roles/tackle/defaults/main.yml
@@ -153,7 +153,6 @@ ui_container_requests_memory: "350Mi"
 ui_ingress_name: "{{ app_name }}"
 ui_ingress_class_name: "nginx"
 ui_ingress_proxy_body_size: "500m"
-ui_ingress_path_type: ImplementationSpecific
 ui_route_name: "{{ app_name }}"
 ui_tls_enabled: false
 ui_tls_secret_name: "{{ ui_service_name }}-serving-cert"
@@ -162,6 +161,10 @@ ui_proto: "{{ 'https' if ui_tls_enabled | bool else 'http' }}"
 ui_node_extra_ca_certs: "/opt/app-root/src/ca.crt"
 ui_route_tls_termination: "edge"
 ui_route_tls_insecure_termination_policy: "Redirect"
+# ui_ingress_path_type should only be defined if you need to override
+# the default value for the ingress controller you are using
+#ui_ingress_path_type: 
+
 
 oauth_provider: openshift
 oauth_default_openshift_sar: --openshift-sar={"namespace":"{{ app_namespace }}","resource":"services","resourceName":"{{ ui_service_name }}","verb":"get"}
--- a/roles/tackle/defaults/main.yml
+++ b/roles/tackle/defaults/main.yml
@@ -146,8 +146,14 @@ ui_container_limits_cpu: "500m"
 ui_container_limits_memory: "800Mi"
 ui_container_requests_cpu: "100m"
 ui_container_requests_memory: "350Mi"
+# Setting ingress defaults for 'nginx' ingress
+# To use 'alb' ingress need to set:
+#   ui_ingress_class_name: "alb"
+#   ui_ingress_path_type: "Prefix"
 ui_ingress_name: "{{ app_name }}"
+ui_ingress_class_name: "nginx"
 ui_ingress_proxy_body_size: "500m"
+ui_ingress_path_type: ImplementationSpecific
 ui_route_name: "{{ app_name }}"
 ui_tls_enabled: false
 ui_tls_secret_name: "{{ ui_service_name }}-serving-cert"
--- a/roles/tackle/templates/ingress-ui.yml.j2
+++ b/roles/tackle/templates/ingress-ui.yml.j2
@@ -3,7 +3,12 @@ apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   annotations:
+{% if ui_ingress_class_name == 'nginx' %}
     nginx.ingress.kubernetes.io/proxy-body-size: {{ ui_ingress_proxy_body_size }}
+{% elif ui_ingress_class_name == 'alb' %}
+    alb.ingress.kubernetes.io/target-type: ip
+    alb.ingress.kubernetes.io/scheme: internet-facing
+{% endif %}
   name: {{ ui_ingress_name }}
   namespace: {{ app_namespace }}
   labels:
@@ -12,14 +17,15 @@ metadata:
     app.kubernetes.io/part-of: {{ app_name }}
     app: {{ app_name }}
 spec:
-  ingressClassName: nginx
+  ingressClassName: {{ ui_ingress_class_name }}
+  
   tls:
     - {}
   rules:
     - http:
         paths:
           - path: /
-            pathType: ImplementationSpecific
+            pathType: {{ ui_ingress_path_type }}
             backend:
               service:
                 name: {{ ui_service_name }}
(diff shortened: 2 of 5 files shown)


roles/tackle/defaults/main.yml
```diff
@@ -144,8 +144,14 @@ ui_container_limits_cpu: "500m"
...
 ui_container_requests_cpu: "100m"
 ui_container_requests_memory: "350Mi"
+# Setting ingress defaults for 'nginx' ingress
+# To use 'alb' ingress need to set:
+#   ui_ingress_class_name: "alb"
+#   ui_ingress_path_type: "Prefix"
 ui_ingress_name: "{{ app_name }}"
+ui_ingress_class_name: "nginx"
 ui_ingress_proxy_body_size: "500m"
+ui_ingress_path_type: ImplementationSpecific
```
jmontleon, line 154: You could do something like this so that the only value that has to be changed is the class name. If that doesn't cover some other unknown scenario it should remain possible to override it from the CR since extra vars have higher precedence than defaults here.
```
ui_ingress_path_type: "{{ 'Prefix' if ui_ingress_class_name == 'alb' else 'ImplementationSpecific' }}"
```

If Prefix/ImplementationSpecific combined with alb/nginx are the only two valid values it could even be written one way or the other (and expanded for other cases as well) with a conditional in the template, but I am fine with it either way.

From experiences with mig-controller, I don't think most people know they can look at defaults/main.yml when looking for options to change behavior, so I don't know if they'll see the comment to change both together (and we probably need to document that alb is an option somewhere (but I don't know if we have any actively updated documentation so that can be a separate PR as far as I am concerned).

roles/tackle/templates/ingress-ui.yml.j2
```diff
@@ -3,7 +3,12 @@ apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   annotations:
+{% if ui_ingress_class_name == 'nginx' %}
     nginx.ingress.kubernetes.io/proxy-body-size: {{ ui_ingress_proxy_body_size }}
+{% elif ui_ingress_class_name == 'alb' %}
+    alb.ingress.kubernetes.io/target-type: ip
+    alb.ingress.kubernetes.io/scheme: internet-facing
+{% endif %}
```
jmontleon, line 11: With two sets of annotations like this the default ansible k8s behavior when updating the value will be to add the new ones, without deleting the old.

I don't know if that will cause any interference, and I'd imagine in practice that won't happen often other than immediately after initial setup, realizing they can't access the UI, and changing it to the other.

Either way to prevent problems we can change the merge_type to merge on creation of this resource.

https://github.com/konveyor/tackle2-operator/blob/main/roles/tackle/tasks/main.yml#L426-L430

https://docs.ansible.com/ansible/latest/collections/kubernetes/core/k8s_module.html#parameter-merge_type

roles/tackle/defaults/main.yml
jwmatthews: Open to doing this now, I imagine in future we will have other Ingresses that need to be supported but I don't know what they would be today.

I imagine at some point we may need to open up the knobs for these various tweaks or defer to the user to create their own Ingress.   I'm not sure how manageable it is for the Operator to be aware of a growing number of Ingress and the tweaks needed.

I could see going simple for now, just have 'nginx' and 'alb' and come back to this later when more are known.

Thoughts?

roles/tackle/templates/ingress-ui.yml.j2
jwmatthews: thank you, I will come back to update the merge_type as you recommend

roles/tackle/defaults/main.yml
jmontleon: Ya, I think if the number grows it get ugly doing it in defaults/main.yml, which is why I brought up the option of doing it in the template.

We should be able to do it in the template with conditionals and make it less ugly (and even make it possible to override in the event one of the configurations we don't know about is somehow possible just by changing this).

[Here](https://github.com/konveyor/tackle2-operator/pull/197/files#diff-bfde856d391d2cee80565372bec171ad04a4efa703c99c53ba92b8b790cfb30dR28)

you could do something like this, not define ui_ingress_path_type in defaults/main.yml then they can flip between alb and nginx with one var. If there are other cases covered by setting ui_ingress_path_type that can be done and/or we can add them to the conditional.

```
{% if ui_ingress_path_type is defined %}
            pathType: {{ ui_ingress_path_type }}
{% elif ui_ingress_class_name == 'alb' %}
            pathType: Prefix
{% else %}
            pathType: ImplementationSpecific
{% endif %}
```

jwmatthews:
@jmontleon this is ready for another review when you have time.

I confirmed we are good with the ALB on EKS with auth enabled.

jmontleon (approved):
ACK

PR #211, opened 2023-04-25
:ghost: Revert catalogsource ns change. It will not work for k8s

jmontleon:


jwmatthews (approved):
Tested this on an EKS cluster and confirmed working.

Thanks @jmontleon

PR #213, opened 2023-04-26
:ghost: Update for konveyor-operator name change

jwmatthews:
Also added ability to change the ingress class name


jmontleon (approved):
ACK

PR #214, opened 2023-04-28
:sparkles: Make rwx_supported false by default

dymurray:
Default behavior should be to not use RWX since it's least common.


jwmatthews (approved):
Agree we want the behavior of rwx to be false so our default is most flexible for all environments

PR #224, opened 2023-05-11
:sparkles: Workaround RHSSO bug by renaming the postgresql service

jmontleon:


dymurray (approved):
Worked for me testing install and upgrade

PR #247, opened 2023-07-05
:seedling: Publish from branch

djzager:
When pushing our manifests to k8s/openshift community-operators, we should do so from the release branch instead of `main`.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -42,7 +42,7 @@ jobs:
           - konveyor/tackle2-ui
           - konveyor/tackle2-addon
           - konveyor/tackle2-addon-windup
-          - konveyor/tackle2-operator
+          - konveyor/operator
           - konveyor/tackle-pathfinder
           - konveyor/tackle-keycloak-theme
     uses: konveyor/release-tools/.github/workflows/create-release.yml@main
@@ -64,7 +64,7 @@ jobs:
           - konveyor/tackle2-ui
           - konveyor/tackle2-addon
           - konveyor/tackle2-addon-windup
-          - konveyor/tackle2-operator
+          - konveyor/operator
           - konveyor/tackle-pathfinder
           - konveyor/tackle-keycloak-init
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/create-release.yml
```diff
@@ -64,7 +64,7 @@ jobs:
           - konveyor/tackle2-ui
           - konveyor/tackle2-addon
           - konveyor/tackle2-addon-windup
-          - konveyor/tackle2-operator
+          - konveyor/operator
```
djzager, line 67: I guess eventually we'll need to update our images to be `quay.io/konveyor/operator:${TAG}`? For now,[looking at our multi-arch build workflow](https://github.com/konveyor/operator/blob/392aa70e7e074e85e26493d78e3524dc6f9e86be/.github/workflows/march-image-build-push.yml#L28C31-L28C31), we should be good living image names alone.

jwmatthews (approved):

.github/workflows/create-release.yml
jmontleon: These changes aren't sufficient if you're going to change the image name here IIUC.
https://github.com/konveyor/operator/pull/247/files/b51f88755b5a44d7588ceecfb4f1a4596da712e3#diff-8ac33fe295df086d3e55df1eb01194819d34cc0f5f54076dba96ceac0e40bd79R67

The script just keys off of the CSV value for the image name:
https://github.com/konveyor/operator/blob/main/tools/konveyor-operator-publish-commands.sh#L135

You need to update all instances of tackle2-operator in:
https://github.com/konveyor/operator/blob/main/Makefile and https://github.com/konveyor/operator/blob/main/config/manifests/bases/konveyor-operator.clusterserviceversion.yaml

Then run `make bundle`. I think that will update to use the new image location.

jmontleon: The build task is still pushing to tackle2-operator but the above is causing the release task to look for the operator image:

https://github.com/konveyor/operator/actions/runs/5469633462/jobs/9958787646

PR #248, opened 2023-07-06
:seedling: Keep operator name the same

djzager:
Even though the project has changed to konveyor/operator, the image we push to quay (at least for now) is konveyor/tackle2-operator.


jmontleon (approved):
ACK. I think this should fix it.

Issue #133, opened 2022-10-17
Provide details when `install-tackle.sh` fails


djzager:
The install-tackle GitHub action will fail ([a link-able example](https://github.com/mundra-ankur/tackle2-hub/actions/runs/3267813719/jobs/5373470530)). We should update the action to dump more diagnostic/debug info in the case of failures, some idea:

1. `kubectl describe pods` - would help if pods wouldn't start
2. `kubectl get subscription` - show olm subscription info

Want to avoid dumping too much to stdout, instead try and narrow down on what the specific failure could be.

djzager:
We should be using the install-konveyor script. I'm going to close this as a won't fix.

PR #252, opened 2023-07-24
:seedling: Update install script with debug output

aufi:
The Tackle/Konveyor install script has few waiting loops checking the installation progress/completeness. Based on last few errors, adding an output with tackle details when waiting loop timeouted to help with debugging CI errors directly in github workflow execution.

Example with recent installation error:
```
...
error: timed out waiting for the condition on tackles/tackle
...
status:
  conditions:
  - lastTransitionTime: "2023-07-24T08:51:07Z"
    message: ""
    reason: ""
    status: "False"
    type: Successful
  - ansibleResult:
      changed: 0
      completion: 2023-07-24T09:05:50.668218
      failures: 1
      ok: 14
      skipped: 48
    lastTransitionTime: "2023-07-24T09:05:51Z"
    message: |
      The task includes an option with an undefined variable. The error was: 'hub_seed_configmap_name' is undefined

      The error appears to be in '/opt/ansible/roles/tackle/tasks/main.yml': line 406, column 3, but may
      be elsewhere in the file depending on the exact syntax problem.

      The offending line appears to be:


      - name: "Setup Hub API Deployment"
        ^ here
    reason: Failed
    status: "False"
    type: Failure
...
```

Related to https://github.com/konveyor/operator/issues/133

### DIFF
--- a/hack/install-tackle.sh
+++ b/hack/install-tackle.sh
@@ -103,7 +103,7 @@ kubectl wait \
   tackles.tackle.konveyor.io/tackle \
 || kubectl get \
   --namespace konveyor-tackle \
-  -o yaml
+  -o yaml \
   tackles.tackle.konveyor.io/tackle # Print debug output when waiting failed
 
 # Now wait for all the tackle deployments
--- a/hack/install-tackle.sh
+++ b/hack/install-tackle.sh
@@ -70,7 +70,8 @@ spec:
   sourceNamespace: konveyor-tackle
 EOF
 # If on MacOS, need to install `brew install coreutils` to get `timeout`
-timeout 600s bash -c 'until kubectl get customresourcedefinitions.apiextensions.k8s.io tackles.tackle.konveyor.io; do sleep 30; done'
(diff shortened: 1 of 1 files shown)


djzager:
This looks like it covers at least a portion of https://github.com/konveyor/operator/issues/133; do you think you could check to see if you could cover whatever remains of that issue?

hack/install-tackle.sh
```diff
@@ -100,12 +101,22 @@ kubectl wait \
...
 
 # Now wait for all the tackle deployments
 kubectl wait \
   --namespace konveyor-tackle \
   --selector="app.kubernetes.io/part-of=tackle" \
   --for=condition=Available \
   --timeout=600s \
-  deployments.apps
+  deployments.apps \
+|| kubectl get \
```
aufi, line 117: Considered using kubectl describe, but that doesn't support filtering items with field-selector.

djzager (approved):

djzager:
Thank you

PR #257, opened 2023-08-08
:seedling: Sequence release of base images

djzager:
Need to build certain images before the rest. This will enable us, in the future when we lock down images for release, to make sure that `tackle2-addon-analyzer` for example has all the appropriate bug fixes from `analyzer-lsp` without pulling from `main`.


jmontleon (approved):
ACK

pranavgaikwad (commented):
Order looks good to me

pranavgaikwad (approved):

PR #262, opened 2023-08-28
:seedling: use previous version when creating release

djzager:
Since we usually provide the "previous_version" as input to the workflow, we should just pass it along to the create release action to prevent us from having to be clever.


dymurray (approved):
LGTM

PR #263, opened 2023-09-15
:sparkles: Allow pathfinder to be optional

fbladilo:
Add feature_pathfinder by default enabled but allow to disable deployment, ds builds do not include pathfinder.


jmontleon (approved):
ACK

PR #266, opened 2023-10-04
Add port forwarding information to main readme

jmle:
I had to find this for my local Linux minikube installation [here](https://github.com/konveyor/operator/blob/main/docs/installation-macos.md), so it probably makes sense to have it in the main README?


jortel:
I'm curious about why this is needed. I run using minikube all of the time and have not had to do this.  I just point my browser at the `minikube ip` to access the UI.

jmle:
@jortel that actually works, I supposed that access to the site was through `localhost`. I can either close this or add the `minikube ip` info

rromannissen:
I think adding the `minikube ip` command makes more sense.

jmontleon (approved):
ACK / LGTM

jmontleon:
Can you resolve the conflicts with this and we'll get it merged. Thank you!

PR #267, opened 2023-10-04
:sparkles: Pass proxy configuration through to Hub pod

mansam:
Intended to resolve https://issues.redhat.com/browse/MTA-849


jmontleon:
Looks like we have `merge_type: merge` on the task already, so if values are removed they should be cleaned up.

LGTM / ACK

jmontleon (approved):

PR #269, opened 2023-10-18
:seedling: Add Pathfinder Assessments migration script

aufi:
Since Pathfinder component is being replaced by Konveyor Hub API, already existing Assessments need to be migrated. Adding a script based on Hub's CLI tool https://github.com/konveyor/tackle2-hub/tree/main/hack/tool that connects to running Pathfinder and migrates Assessments to running Hub API v0.3.

This script is expected to be executed by operator during upgrade Konveyor 0.2->0.3.

Example usage: (only auth token argument is optional, both urls are required)

```
$ ./migrate-pathfinder-assessments --pathfinder-url http://pathfinder.svc --hub-base-url http://hub.svc:8080 --token ""
```

Related to https://github.com/konveyor/tackle2-hub/issues/489 and https://github.com/konveyor/tackle2-hub/pull/507

### DIFF
--- a/tools/upgrades/migrate-pathfinder-assessments.py
+++ b/tools/upgrades/migrate-pathfinder-assessments.py
@@ -73,11 +73,11 @@ def migrateAssessments(pathfinder_url, hub_base_url, token):
     print("There are %d Applications, looking for their Assessments.." % len(apps))
     for app in apps:
         # If there would be more assessments, only first one is migrated.
-        for passmnt in apiJSON(pathfinder_url + "/assessments?applicationId=%d" % app.id, token):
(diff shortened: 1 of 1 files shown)


jmontleon (approved):
ACK

aufi:
@jmontleon Thanks for ack, I'm going to merge the upgrade script to operator. At some point we might add the whole tackle CLI tool here, but this should be enough for assessments migration (and doesn't bring any new dependencies like the full tool version).

jmontleon:
Thank you! I'll work on a PR to detect Pathfinder, run the Migration, and take Pathfinder down.

PR #270, opened 2023-10-23
:seedling: use non-docker.io image for oauth proxy

djzager:


pranavgaikwad (approved):

kaovilai:
weird `bundle/` isn't updated

kaovilai (changes requested):
`bundle/` should probably reflect the change too. `make bundle` working?

jortel (approved):
LGTM

PR #274, opened 2023-10-24
:seedling: simplify pr checks

djzager:


.github/workflows/csv-validate.yml
```diff
@@ -1,13 +1,12 @@
...
 on:
   pull_request:
     branches:
     - main
-    - 'release-*.*.*'
+    - 'release-*.*'
     paths:
-    - 'bundle/**'
-    workflow_dispatch:
+    - 'config/**'
```
kaovilai, line 9:
```suggestion
    - 'config/**'
    - 'bundle/**'
```
might make sense to check both.. since one might accidentally only update `bundle/` and not `config`

kaovilai (approved):

dymurray (approved):

PR #275, opened 2023-10-27
:seedling: switch from kustomize to helm

djzager:
attempting to drop-in replace kustomize with helm for easier management of release artifacts


fabianvf:
nice, looks like it gets rid of a lot of cruft too. Does this affect QE at all? Have we done a full test run to see what the impact is?

kaovilai:
probably make sense as a project with barely any go code or CRD generations.

hack/install-konveyor.sh
```diff
@@ -0,0 +1,97 @@
...
+  || kubectl get subscription --namespace ${NAMESPACE} -o yaml konveyor-operator # Print subscription details when timed out
+}
+
+install_tackle() {
+  echo "Waiting for the Tackle CRD to become available"
+  kubectl wait \
+    --namespace ${NAMESPACE} \
+    --for=condition=established \
+    customresourcedefinitions.apiextensions.k8s.io/tackles.tackle.konveyor.io
+
```
kaovilai, line 56: So looking at testdata folder of operator-sdk I think this remains an anti-pattern.

https://github.com/operator-framework/operator-sdk/tree/master/testdata/helm/memcached-operator/config/crd

`config/` folder and kustomize still used. Is the intent to break away from operator-sdk helm pattern?

kaovilai: This PR essentially mean  going back to manually annotating CSV in bundle/ and the main pain point for me is converting into JSON the alm-examples which `config/samples` solves.

https://sdk.operatorframework.io/docs/building-operators/helm/quickstart/

kaovilai: CRDs are now in `./helm/templates` which is not the case in `https://github.com/operator-framework/operator-sdk/tree/master/testdata/helm/memcached-operator/helm-charts/memcached/templates`.

Just noting the anti-pattern. If everyone is fine with that, then its a non-issue.

Staying in pattern could make future migrations to https://github.com/operator-framework/rukpak (supersede operator-sdk) easier.

djzager: My understanding is that memcached-operator is a helm based operator...meaning the operand is being deployed using some variation of `helm install`. Here we are just trying to make it easier to generate the operator manifests because the customization of `kustomize` manifests is so cumbersome (specifically the images that are inside environment variables `RELATED_IMAGES_`). The other really painful thing is computing and replacing skips and replaces.

Basically, we want to be able to helm install the operator (that is ansible based). I could rename `./helm` to `./config` but I think that may confuse people looking at the project.

Lastly, the only real intelligence I see from operator-sdk is in taking a bunch of YAMLs on stdin and generating the `./bundle` dir. We aren't leveraging any of the controller-gen tooling that would make this more difficult.

kaovilai: > because the customization of kustomize manifests is so cumbersome (specifically the images that are inside environment variables RELATED_IMAGES_

got it.

> memcached-operator is a helm based operator.

They have [helm](https://github.com/operator-framework/operator-sdk/tree/master/testdata/helm/memcached-operator), [go/v3](https://github.com/operator-framework/operator-sdk/tree/master/testdata/go/v3), and [go/v4-alpha](https://github.com/operator-framework/operator-sdk/tree/master/testdata/go/v4-alpha) variants of memcached operator.

> computing and replacing skips and replaces.

If you own the catalogsource, [opm tooling makes it pretty easy](https://github.com/kaovilai/oadp-operator/blob/edge/appendMakefile#L2) to add new bundle as a replacement of current one.

.github/actions/make-bundle/action.yml
```diff
@@ -0,0 +1,65 @@
+name: Make Operator Bundle
+description: |
+  Make an operator bundle and save it. This does not save the image for you.
```
djzager, line 3: I like contradicting sentences next to each other like this.

.github/actions/make-bundle/action.yml
```diff
@@ -0,0 +1,65 @@
...
+inputs: 
+  operator_bundle:
+    description: "image uri for operator bundle"
```
kaovilai, line 6: I would add examples just so people know expected formatting.

bundle/manifests/konveyor-operator.clusterserviceversion.yaml
```diff
@@ -8,8 +8,7 @@ metadata:
           "apiVersion": "tackle.konveyor.io/v1alpha1",
           "kind": "Tackle",
           "metadata": {
-            "name": "tackle",
-            "namespace": "konveyor-tackle"
```
kaovilai, line 12: Is this intentional?

hack/start-minikube.sh
```diff
@@ -2,13 +2,28 @@
...
+#  - ${PROJECT_ROOT}/bin/install-tackle.sh
+__dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+__root="$(cd "$(dirname "${__dir}")" && pwd)"
+__repo="$(basename "${__root}")"
+__bin_dir="${__root}/bin"
+__os="$(uname -s | tr '[:upper:]' '[:lower:]')"
+__arch="$(uname -m | sed 's/x86_64/amd64/' | sed 's/aarch64/arm64/')"
+
+# Update PATH for execution of this script
+export PATH="${__bin_dir}:${PATH}"
```
kaovilai, line 17: If L14 don't need exporting, then why is L17 export needed?

helm/templates/olm/scorecard.yaml
```diff
@@ -1,50 +1,51 @@
...
+kind: Configuration
+metadata:
+  name: config
+stages:
+- parallel: true
+  tests:
+  - entrypoint:
+    - scorecard-test
+    - basic-check-spec
+    image: quay.io/operator-framework/scorecard-test:v1.22.0
```
kaovilai, line 12:
```suggestion
    image: quay.io/operator-framework/scorecard-test:v1.22
```
This probably doesn't hurt :)

bundle/manifests/konveyor-operator.clusterserviceversion.yaml
djzager: It was but I'm not certain now why I did it. Will check this again.

hack/install-konveyor.sh
```diff
@@ -0,0 +1,97 @@
...
+__root="$(cd "$(dirname "${__dir}")" && pwd)"
+__repo="$(basename "${__root}")"
+__bin_dir="${__root}/bin"
+__os="$(uname -s | tr '[:upper:]' '[:lower:]')"
+__arch="$(uname -m | sed 's/x86_64/amd64/' | sed 's/aarch64/arm64/')"
+
+# Update PATH for execution of this script
+export PATH="${__bin_dir}:${PATH}"
+
+NAMESPACE="${NAMESPACE:-konveyor-tackle}"
```
ascheman, line 20: Do you do this on purpose with this syntax? The classical way to perform this would be

```
: "${NAMESPACE:=konveyor-tackle}"
```

hack/install-konveyor.sh
```diff
@@ -0,0 +1,97 @@
...
+#  - ${PROJECT_ROOT}/bin/install-tackle.sh
+__dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+__root="$(cd "$(dirname "${__dir}")" && pwd)"
+__repo="$(basename "${__root}")"
+__bin_dir="${__root}/bin"
+__os="$(uname -s | tr '[:upper:]' '[:lower:]')"
+__arch="$(uname -m | sed 's/x86_64/amd64/' | sed 's/aarch64/arm64/')"
+
+# Update PATH for execution of this script
+export PATH="${__bin_dir}:${PATH}"
```
ascheman, line 18: $PATH is always exported, no need, to re-export it

hack/install-konveyor.sh
```diff
@@ -0,0 +1,97 @@
...
+
+NAMESPACE="${NAMESPACE:-konveyor-tackle}"
+OPERATOR_BUNDLE_IMAGE="${OPERATOR_BUNDLE_IMAGE:-quay.io/konveyor/tackle2-operator-bundle:latest}"
+TACKLE_CR="${TACKLE_CR:-}"
+
+if ! command -v kubectl >/dev/null 2>&1; then
```
ascheman, line 24: Do you want to see if `kubectl` is on the PATH?
I think
`command kubectl 2>&-` is a little shorter and does the job as well

hack/install-konveyor.sh
djzager: Ahh. I see. Reading https://tldp.org/LDP/abs/html/parameter-substitution.html looks like there might be slight difference between:

```
NAMESPACE="${NAMESPACE:-konveyor-tackle}"
NAMESPACE="${NAMESPACE:=konveyor-tackle}"
```

I am not immediately seeing a scenario where one will work and the other will fail to work as intended. Thoughts?

hack/install-konveyor.sh
```diff
@@ -0,0 +1,97 @@
+#!/bin/bash
+
+set -e
+set -x
+
```
ascheman, line 5: Perhaps you also want to add `set -o pipefail`? If a (pipeline) chained command fails, it only breaks the execution if the last command in the pipeline fails.

ascheman:
Taking a deeper look into `hack/install-konveyor.sh` I strongly recommend to run shellcheck (and fix the issues) before merging.

hack/start-minikube.sh
ascheman: The export is not needed here (cf. https://github.com/konveyor/operator/pull/275#discussion_r1380130973). The general difference here is that `${PATH}` is an environment variable (by the export) and `${__arch}` (and the others) are shell variables.

kaovilai: Then keep it unexported so it don't alter user environment?

ascheman: No, the concept is very different: We are on Unix, not on Windows. Environment-Variables are only inherited by new sub-processes of the running process. They never ever affect the environment of any other process (even of the same user). Shell-variables are only relevant to the current shell/script. They are never inherited by new sub-processes (even if the sub-process itself is a shell).

hack/install-konveyor.sh
djzager: I didn't realize this. Will examine.

hack/install-konveyor.sh
djzager: The one provided doesn't handle stdout. Could make `>/dev/null` to `>-` but I think it's fine to have a few more character for clarity.

ascheman: `command kubectl` doesn't print something (only the `-v` flag does), so redirecting stdout to the null device is not necessary. Without `-v` it just changes its exit value depending on the existence of the executable on ${PATH}.

djzager: Am I missing something?

```
bash-3.2$ command kubectl 2>&-
kubectl controls the Kubernetes cluster manager.

 Find more information at: https://kubernetes.io/docs/reference/kubectl/

Basic Commands (Beginner):
  create          Create a resource from a file or from stdin
  expose          Take a replication controller, service, deployment or pod and expose it as a new Kubernetes service
  run             Run a particular image on the cluster
  set             Set specific features on objects

...
```

kaovilai (approved):

hack/install-konveyor.sh
ascheman: Sorry, you are completely right, my fault. You have to redirect stdout as well.

ascheman: Sorry for the confusion, I mixed up one thing with the `which` command: `command` without the `-v` flag will even execute the given executable. Thats the reason why you see the standard output, it is just the help text of `kubectl` - be _cautious_: other commands may even have side effects. Hence, you should definitely use `command -v ...` as in the beginning, as it will only check the existence of the executable (and print its path).

ascheman: I was verifying only with `command false` and `command true` which doesn't print something at all (but nevertheless may have side effects, depending on the given exe).

jmontleon (approved):
ACK / LGTM

PR #276, opened 2023-10-31
:sparkles: Migrate Pathfinder assessments and remove Pathfinder

jmontleon:


jmontleon:
Work in progress mostly because something still seems to need sorting out with the actual migration:
```
Starting Pathfinder Assessments to Konveyor Assessment migration.
There are 10 Applications, looking for their Assessments..
# Assessment for Application 83
Traceback (most recent call last):
  File "/usr/local/bin/migrate-pathfinder-assessments.py", line 124, in <module>
    appCnt = migrateAssessments(args.pathfinder_url, args.hub_base_url, args.token)
  File "/usr/local/bin/migrate-pathfinder-assessments.py", line 89, in migrateAssessments
    for sh in passmnt['stakeholders']:
KeyError: 'stakeholders'
```

jmontleon:
cc @djzager for removing the image from GH actions, etc.

roles/tackle/tasks/main.yml
```diff
@@ -519,3 +452,112 @@
...
+
+- name: "Remove PathFinder PostgreSQL Volume"
+  k8s:
+    state: absent
+    api_version: v1
+    kind: PersistentVolumeClaim
+    name: "{{ pathfinder_database_data_volume_claim_name }}"
+    namespace: "{{ app_namespace }}"
+  when:
+    - (pathfinder_delete_db_volume|bool)
```
jmontleon, line 562: I put the DB volume removal behind a flag since there's no going back. To me it seems like this should be a conscious users decision to make once they feel confident everything is as it should be. They can also delete it manually `oc delete pvc ...`.

aufi:
From my side, looks great!

jmontleon:
I tested with a clean install and an upgrade. Both seemed to go fine, and in the case of the upgrade assessments were copied over.

djzager (approved):

tools/templates/clusterserviceversion.yaml.j2
djzager, line 1: My understanding is that the changes to this file were for completeness. I'm not aware of anything we are currently using that relies on this file.

jmontleon: Yes, just trying to remove all references to Pathfinder. I can follow up with a PR to drop the file so we can re-add it easily if we discover something does use it.

PR #277, opened 2023-11-06
:bug: adding ability for the metadata.namespace to be what we watch

shawn-hurley:


djzager (approved):

shawn-hurley:
Will this break any of the CSV generation or something though?

djzager:
you need to run `make bundle` and I need to understand why the "Validate Bundle CSV" check is not failing...

djzager:
Turns out the check isn't failing because `operator-sdk generate bundle` changes the `WATCH_NAMESPACE` for you. All you need to do is sign your commit.

PR #284, opened 2023-11-20
:seedling: Use konveyor/ci in release workflow

djzager:
* Also add new targets to Makefile for installing openshift-client and yq
* Refactor our method for preparing manifests for publish


.github/actions/make-bundle/action.yml
```diff
@@ -41,25 +41,43 @@ inputs:
...
     default: ""
+  version:
+    description: "operator version"
+    required: false
+    default: ""
+  channels:
+    description: "comma separated channel(s) this operator should be available on"
+    required: false
+    default: ""
+  # use_image_digest:
```
fabianvf, line 52: are we leaving this in for a reason?

.github/workflows/create-release.yml
```diff
@@ -131,27 +115,196 @@ jobs:
...
+      operator: quay.io/konveyor/tackle2-operator:${{ inputs.version }}
+      tackle_hub: quay.io/konveyor/tackle2-hub:${{ inputs.version }}
+      keycloack_init: quay.io/konveyor/tackle-keycloak-init:${{ inputs.version }}
+      tackle_ui: quay.io/konveyor/tackle2-ui:${{ inputs.version }}
+      addon_admin: quay.io/konveyor/tackle2-addon:${{ inputs.version }}
+      addon_analyzer: quay.io/konveyor/tackle2-addon-analyzer:${{ inputs.version }}
+
   publish:
     name: Build and Push Manifest
-    needs: release-components
```
fabianvf, line 136: do we want this one to depend on tests or something?

fabianvf (commented):
This looks sane to me, just a few questions

.github/workflows/create-release.yml
djzager: it should. Good find.

.github/actions/make-bundle/action.yml
djzager: For some strange reason when I use the `operator-sdk generate bundle ... --use-image-digests` on our project it doubles up the `relatedImages` where each image is repeated multiple times. I tried to recreate the issue on memcached-operator and couldn't. Not certain it's worth the time.

Dropping for now.

fabianvf (approved):
LGTM

PR #304, opened 2024-02-14
:bug: tackle2-addon is no longer deployed

### LINKED ISSUE #300: Drop tackle2-addon from manifests
We don't need to reference the tackle2-addon container image in our helm chart/operator manifests. This means that we should remove references to it in:

1. operator roles (maybe. Not immediately obvious if this is worth the effort)
2. operator related manifests. presumably most of the occurrences of quay.io/konveyor/tackle2-addon would be a safe bet to remove
3. build pipeline. NOTE: we need to drop the waiting for the image. we still need to cut the release

djzager:
Fixes #300


dymurray (approved):
LGTM

jmontleon (approved):
LGTM

PR #306, opened 2024-03-28
:bug: Update the postgresql image tag.

jmontleon:


ibolton336 (approved):
Tested locally. LGTM
Screenshot 2024-03-28 at 1 59 23 PM

djzager (changes requested):
Just make this a :bug: so that it appears in release notes. Additionally I wonder if this needs to go into release-0.3

jmontleon:
@djzager we mirror the image with SHA for releases don't we?

djzager (approved):

sshveta (approved):

PR #309, opened 2024-04-02
:book: Update OLM version in installation-macos.md

lucamolteni:
v0.21.2 fails on Apple Silicon ARM due to this https://github.com/operator-framework/operator-lifecycle-manager/issues/2823#issuecomment-1468421055

A minimum of 0.24.0 is required, I managed to install it on my machine with 0.27.0

### DIFF
--- a/docs/installation-macos.md
+++ b/docs/installation-macos.md
@@ -16,7 +16,7 @@ podman machine start
 3. Install ingress addon
 `minikube addons enable ingress`
 4. Install OLM to manage Konveyor operator
-`curl -sL https://github.com/operator-framework/operator-lifecycle-manager/releases/download/v0.21.2/install.sh | bash -s v0.21.2`
+`curl -sL https://github.com/operator-framework/operator-lifecycle-manager/releases/download/v0.27.0/install.sh | bash -s v0.27.0`
 5. Install Konveyor operator
 `kubectl create -f https://operatorhub.io/install/konveyor-operator.yaml`
 6. Verify if the Konveyor operator pod is running or not


lucamolteni:
@jmontleon can you check this? I still wasn't able to complete an installation on konveyor on an Silicon Mac due to 

https://kubernetes.slack.com/archives/CR85S82A2/p1712153667784579

```
- name: tackle-pathfinder-postgresql
image: >-
quay.io/konveyor/postgresql-12-centos7@sha256:1287d947464fbf5275feb31d106d2feb856c6117f0eaec8d947f2e03bd8618c5
 https://hub.docker.com/r/centos/postgresql-12-centos7/tags
```

This image is running the wrong architecture, here's the error

```
exec /usr/bin/container-entrypoint: exec format error
```

jmontleon:
It doesn't look like the centos7 container is multiarch.

This is where our image is coming from; we copy it to our org to guarantee the SHA will remain and then refer to it by SHA:
https://quay.io/repository/centos7/postgresql-12-centos7?tab=tags

I have a PR up that probably needs more testing to update us to postgresql 15, which does look like it has ARM images up
https://quay.io/repository/sclorg/postgresql-15-c9s?tab=tags

https://github.com/konveyor/operator/pull/310

lucamolteni:
> It doesn't look like the centos7 container is multiarch.
> 
> This is where our image is coming from; we copy it to our org to guarantee the SHA will remain and then refer to it by SHA: https://quay.io/repository/centos7/postgresql-12-centos7?tab=tags
> 
> I have a PR up that probably needs more testing to update us to postgresql 15, which does look like it has ARM images up https://quay.io/repository/sclorg/postgresql-15-c9s?tab=tags
> 
> #310

This seems to be really promising thank you

I'd like to try it on my machine so that I check if it works or not, but it seems like the build process requires the access to a quay.io repository. Any ideas on how to try it quicker?

lucamolteni:
> Do you just require the `quay.io/konveyor/postgresql-15-c9s` image to exist?
> 
> If so I can skopeo copy it over so you can test.

I'm not sure, but I think that to try your PR I should recompile the operator, I was trying to use the `install-konveyor.sh` script in ~/hack but it runs

`docker build --platform linux/amd64 -t quay.io/konveyor/tackle2-operator:latest`

jmontleon:
Most of the changes in my PR aside from actually changing the image are to handle upgrades from 12 to 15. You shouldn't really need any of that to run with a newer version.

Looking at the instructions you're following, it looks like you did:
`kubectl create -f https://operatorhub.io/install/konveyor-operator.yaml`

The only thing we use postgres for now is keycloak if you enable auth. If you download this file and edit the subscription to `channel: konveyor-0.3` you should be able to run without postgres/auth.
```
---
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  name: my-konveyor-operator
  namespace: my-konveyor-operator
spec:
  channel: konveyor-0.1
  name: konveyor-operator
  source: operatorhubio-catalog
  sourceNamespace: olm
```

If you want to override the postgres image you can override the image in the tackle CR.
```
spec:
  keycloak_database_image_fqin: quay.io/konveyor/postgresql-15-c9s
```
But you're going to hit the same issue with the keycloak container. 
https://quay.io/repository/konveyor/keycloak?tab=tags

I think someone (@fbladilo) is looking at updating keycloak to a more current version, where we'd at least get arm builds.

But until that happens you probably just want to update the channel and keep auth disabled, sidestepping the problem.

jmontleon (approved):

PR #316, opened 2024-06-03
:ghost: Add Architecture support labels

jmontleon:
I have admittedly not tested extensively, but we can install and the UI comes up. Dependencies are also far less problematic on arm64 than ppc64le and s390x. However, the operator does not show up on arm64 clusters, I _think_ because we are missing these labels.

### DIFF
--- a/bundle/manifests/konveyor-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/konveyor-operator.clusterserviceversion.yaml
@@ -20,7 +20,7 @@ metadata:
     categories: Modernization & Migration
     certified: "false"
     containerImage: quay.io/konveyor/tackle2-operator:latest
-    createdAt: "2024-02-14T17:22:48Z"
+    createdAt: "2024-06-03T20:19:41Z"
     description: Konveyor is an open-source application modernization platform that
       helps organizations safely and predictably modernize applications to Kubernetes
       at scale.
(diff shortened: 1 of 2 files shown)


jmontleon:
tests are failing because centos stream is eol and they've apparently already pulled the repos. I will look to see if I can find an alternative.

shawn-hurley (approved):
do we have things set up to make sure the correct things are downstream too? Otherwise works for me

@fbladilo

jmontleon:
@shawn-hurley Good point. I'm not sure what we want to do for downstream in the immediate future. Downstream we just might want to be aware so this can be patched out. @fbladilo @rayfordj does that seem reasonable.

kaovilai (approved):

bundle/manifests/konveyor-operator.clusterserviceversion.yaml
```diff
@@ -42,6 +42,11 @@ metadata:
     operators.operatorframework.io/project_layout: ansible.sdk.operatorframework.io/v1
     repository: https://github.com/konveyor/tackle2-operator
     support: https://github.com/konveyor/tackle2-operator/issues
+  labels:
+    operatorframework.io/arch.amd64: supported
+    operatorframework.io/arch.arm64: supported
```
rayfordj, line 47:
```suggestion
    operatorframework.io/arch.arm64: unsupported
```

helm/templates/olm/konveyor-operator.clusterserviceversion.yaml
```diff
@@ -27,6 +27,11 @@ metadata:
     operatorframework.io/suggested-namespace: konveyor-tackle
     repository: https://github.com/konveyor/tackle2-operator
     support: https://github.com/konveyor/tackle2-operator/issues
+  labels:
+    operatorframework.io/arch.amd64: supported
+    operatorframework.io/arch.arm64: supported
```
rayfordj, line 32:
```suggestion
    operatorframework.io/arch.arm64: unsupported
```

rayfordj (changes requested):

helm/templates/olm/konveyor-operator.clusterserviceversion.yaml
jmontleon: @rayfordj we want arm64 supported upstream.

rayfordj (approved):
/lgtm

Clarified in Slack chat -- upstream to build for ARM; downstream to continue as-is.

PR #317, opened 2024-06-04
:ghost: don't wait for rulesets to build

djzager:
because it never will


dymurray (approved):

djzager:
Merging because #318 can address the build failures.

PR #320, opened 2024-06-04
:bug: Use AlmaLinux 8 instead of CentOS 8 Stream Vault

jmontleon:
Switch to Alma. The CentOS 8 Stream Vault metadata seems to be broken. The postgres:15 module is unavailable and builds are still breaking. I've beem locally with Alma and it works. They also support all architectures for AlmaLinux 8. Rocky and others are lacking, at least according to this chart:
https://wiki.almalinux.org/Comparison.html


djzager (approved):
lol. yay

PR #324, opened 2024-06-04
:ghost: Add CSV annotations

jmontleon:
I am not certain about this one. Can others weigh in:

```
features.operators.openshift.io/proxy-aware

Specify whether an Operator supports running on a cluster behind a proxy by accepting the standard HTTP_PROXY and HTTPS_PROXY proxy environment variables. If applicable, the Operator passes this information to the workload it manages (operands).
```

### DIFF
--- a/bundle/manifests/konveyor-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/konveyor-operator.clusterserviceversion.yaml
@@ -24,6 +24,16 @@ metadata:
     description: Konveyor is an open-source application modernization platform that
       helps organizations safely and predictably modernize applications to Kubernetes
       at scale.
+    features.operators.openshift.io/cnf: "false"
+    features.operators.openshift.io/cni: "false"
+    features.operators.openshift.io/csi: "false"
+    features.operators.openshift.io/disconnected: "true"
+    features.operators.openshift.io/fips-compliant: "false"
+    features.operators.openshift.io/proxy-aware: "true"
+    features.operators.openshift.io/tls-profiles: "false"
+    features.operators.openshift.io/token-auth-aws: "false"
+    features.operators.openshift.io/token-auth-azure: "false"
+    features.operators.openshift.io/token-auth-gcp: "false"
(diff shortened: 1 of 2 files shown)


rayfordj (approved):
/lgtm

Thanks!

bundle/manifests/konveyor-operator.clusterserviceversion.yaml
```diff
@@ -20,10 +20,20 @@ metadata:
...
-    createdAt: "2024-06-03T20:19:41Z"
+    createdAt: "2024-06-06T13:28:50Z"
     description: Konveyor is an open-source application modernization platform that
       helps organizations safely and predictably modernize applications to Kubernetes
       at scale.
+    features.operators.openshift.io/cnf: "false"
+    features.operators.openshift.io/cni: "false"
+    features.operators.openshift.io/csi: "false"
+    features.operators.openshift.io/disconnected: "true"
+    features.operators.openshift.io/fips-compliant: "false"
```
shawn-hurley, line 31: Why are we not designed for fips? Maybe we just need to do the exercise to make sure, but I think we should. @dymurray thoughts?

shawn-hurley (approved):
I have a minor concern that is probably outside the scope of this PR. 

We should just do that as a follow on if we decide we are actually fine

bundle/manifests/konveyor-operator.clusterserviceversion.yaml
rayfordj: @shawn-hurley , I don't know about "designed for", but we've not yet configured downstream to build to run on FIPS enabled clusters.  Until that is done, we should continue to claim "false" for its value.  Franco has a task configure and build MTA (7.0.z & 7.1.z) for it , but it is lower priority than other efforts around MTA at the moment.

rayfordj: FWIW, we're already injecting this in downstream CSV.  This PR just brings the change more visibility and the two CSVs closer parity

rayfordj (approved):

PR #329, opened 2024-06-06
:seedling: double the ttl on operator image

djzager:
This should make "rerunning failed" tasks in workflows less likely to fail simply because the image didn't exist anymore.


jmontleon (approved):
ACK

PR #336, opened 2024-06-17
:sparkles: Update CRDs to v1alpha2

jmontleon:


jmontleon:
@shawn-hurley can you help verify this is correct for adding a new CRD version?

I'm a little iffy. According to https://docs.openshift.com/container-platform/4.15/operators/operator_sdk/osdk-generating-csvs.html 

We should `Add a new entry in the CRD resource under the versions section of your CSV.`  There is no versions section of a CSV that I am aware of. There is `spec.versions[]` in the CRD with served and storage that I believe they're talking about.

And then `Ensure the referencing version of the CRD in the owned section of your CSV is updated if the CSV intends to use the new version:`. 

I guess we just list the new version?

jmontleon:
The tests are failling from:
```
 TASK [Remove Admin Addon CR] ******************************** 
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Failed to find exact match for tackle.konveyor.io/v1alpha1.Addon by [kind, name, singularName, shortNames]"}
```

This is fixed by:
https://github.com/konveyor/operator/pull/336/files#diff-ac6f51e564930b0b25ed41f9a0d935bc87f23ab0ae54de6f725b7065f043f9f9R8

But tests will not pass until everything is merged because the bundle is still using image: `quay.io/konveyor/tackle2-operator:latest`.

@djzager fyi we have an issue where operator role changes don't get tested. We probably need to build the container image replace it when building the bundle for tests to properly account for changes.

jortel (approved):
LGTM

PR #340, opened 2024-06-18
:bug: build the bundle before build+push

### LINKED ISSUE #338: pr-ci.yml does not build / use an operator container that contains PR
https://github.com/konveyor/operator/pull/333 had a change to the operator role that was required to successfully install, but the bundle specifies the published container image in the konveyor org.

I think we need to build the operator container image and use it in the generated bundle.

djzager:
Fixes #338


jmontleon (approved):
LGTM! Thank you!

PR #347, opened 2024-06-27
:seedling: pin provider images in our release artifacts

djzager:


.github/workflows/create-release.yml
```diff
@@ -267,11 +269,6 @@ jobs:
     uses: konveyor/ci/.github/workflows/global-ci-bundle.yml@main
     with:
       operator_bundle: quay.io/konveyor/tackle2-operator-bundle:${{ inputs.version }}
-      tackle_hub: quay.io/konveyor/tackle2-hub:${{ inputs.version }}
```
djzager, line 270: We don't need these because they should already be defined in the bundle

jmontleon (approved):
ACK

PR #355, opened 2024-07-10
:bug: Fix hard coded namespace in ansible role task

jmontleon:


jmontleon:
This fails with MTA because I hard coded the namespace value.

dymurray (approved):

PR #360, opened 2024-07-18
:seedling: wait for olm

djzager:


hack/install-konveyor.sh
```diff
@@ -92,5 +92,6 @@ EOF
 }
 
 kubectl get customresourcedefinitions.apiextensions.k8s.io clusterserviceversions.operators.coreos.com || operator-sdk olm install
+kubectl wait --for=condition=established customresourcedefinitions.apiextensions.k8s.io clusterserviceversions.operators.coreos.com
```
aufi, line 95: Thinking about timeout like 600s, but not a blocker.

aufi (approved):

PR #366, opened 2024-07-23
:bug: reducing requests and limits after we have better control of the JVM

shawn-hurley:
related to https://github.com/konveyor/java-analyzer-bundle/pull/106


jmontleon (approved):
ACK

PR #370, opened 2024-07-23
:sparkles: Increase operator limits

jmontleon:


jmontleon:
Our operators reconcile loop is painfully slow. Increasing resources to speed it up some.

PR #373, opened 2024-08-01
:sparkles: Adjust resource requirements.

jortel:
The resource requirements for the analyzer and providers can be lowered.  This will support better scaling.  Also, for the java provider, the JDK will not use 2Gi.

Scale testing has shown the hub memory spike to 2.3Gi with 3k applications being analyzed as fast as possible.  Recommend limit should be _bumped_.

**NOTE**:  I have some real concerns about large differences between request/limit.  I'm concerned about oversubscribing the nodes when launching hundreds/thousands of pods without a Resource Quota.  k8s does not seem to protect itself (nodes) against this.

Can the builtin requirements be lower?


roles/tackle/defaults/main.yml
```diff
@@ -176,7 +176,7 @@ provider_java_component_name: "extension"
 provider_java_container_limits_cpu: "1"
 provider_java_container_limits_memory: "2Gi"
 provider_java_container_requests_cpu: "1"
-provider_java_container_requests_memory: "2Gi"
+provider_java_container_requests_memory: "1Gi"
```
shawn-hurley, line 179: This should stay at 2GI for the request the JVM will use 70% of max memory, and means that we can assume that it will use the full amount IMO. It will be safer to set this limit equal to the requests

jortel:
Looks like requested resources have already been adjusted.

PR #385, opened 2024-10-24
:bug: Set the maven local repository path.

jortel:
Following the _AnyUser_ change, the local m2 repository path needs to be set to /cache/m2 so the java provider will not use it's default path of /root/.m2.


dymurray (approved):

dymurray:
merging because CI is broken with the latest OLM

PR #395, opened 2024-11-07
:bug: Update PVC permissions

jmontleon:


shawn-hurley (approved):
This looks right to me, VISACK

jmontleon:
For whatever reason, when the cache PVC exists the Hub mounts it, so we must scale it down to allow the PVC to terminate so it can be recreated.

jmontleon:
Thinking about this more, since the Hub mounts the cache PVC if it exists we probably don't need to delete and recreate it. We should be able to chown everything like the others. It could however take awhile for the process to run, whereas deleting the cache and recreating it may be quicker. Curious what others opinions are with which way we should go.

shawn-hurley:
I think that deleting the cache should be fine and as you said would be quicker

roles/tackle/tasks/main.yml
```diff
@@ -3,6 +3,15 @@
...
     api_groups: "{{ lookup('k8s', cluster_info='api_groups') }}"
 
+- name: Gather Facts
+  ansible.builtin.gather_facts:
+    gather_subset:
+      - user_uid
+
+- name: "Retrieve UID"
+  set_fact:
+    hub_task_uid: "{{ ansible_facts['user_uid'] }}"
```
jortel, line 13: This is the hub UID, not the task UID, correct?  Perhaps `hub_uid` would be clearer?

jmontleon: To be clear, when we retrieve this value we are getting it operator pod. The operator pod does not run as root, so it uses the default namespace UID.
```
$ oc exec -it mta-operator-cdf4858b7-7wch7 /bin/bash
kubectl exec [POD] [COMMAND] is DEPRECATED and will be removed in a future version. Use kubectl exec [POD] -- [COMMAND] instead.
bash-4.4$ id
uid=1000860000(1000860000) gid=0(root) groups=0(root),1000860000
```

We then set the hub to application container to run with this UID:
https://github.com/konveyor/operator/pull/395/files#diff-9e3528a228e3b18c37a6c6689149af80bcd315d76a2a3c2c0dd920e87fd6e07eR208

And pass it to the init container (which runs as root) so it can chown the files to the same UID:
https://github.com/konveyor/operator/pull/395/files#diff-9e3528a228e3b18c37a6c6689149af80bcd315d76a2a3c2c0dd920e87fd6e07eR223-R234

And then it gets passed down to the task pod as well (if not overridden with the now available environment variable).

I was trying to convey that this is the UID that will be used for the hub and task pods, but am open to changing the var name to something clearer/better if you have something in mind.

jmontleon: didn't see `hub_uid` suggestion first time reading. Updated it to use this.

jmontleon:
Fixed the hard coded values in the initContainer causeing the e2e-api test failure.

PR #409, opened 2025-03-14
:bug: Fix oauth to work with external access networkpolicy

jmontleon:
The external access network policy allows 8080 and 8443. So using 8443 instead of 9443 allows everything to work.


rayfordj (approved):
/approve

PR #412, opened 2025-03-17
:sparkles: adding just auth to the install of tackle

shawn-hurley:
Follow on to only add the auth to the install of tackle to actually test login in UI tests.


jmontleon (approved):
ACK

PR #416, opened 2025-03-19
:bug: disable keycloak servicemonitor

jmontleon:
The reason for this is that there is a bug that causes the metrics/monitoring target to be broken and return a 404. Disabling it and cleaning up the servicemonitor is the best we can do to stop the noise right now. With RHBK there is no servicemonitor installed so nothing to do there.
https://issues.redhat.com/browse/RHSSO-1960


djzager (approved):
lgtm

PR #418, opened 2025-03-19
:bug: Fix oauth to work with external access networkpolicy

jmontleon:


roles/tackle/defaults/main.yml
```diff
@@ -152,7 +152,7 @@ oauth_provider: openshift
 oauth_default_openshift_sar: --openshift-sar={"namespace":"{{ app_namespace }}","resource":"services","resourceName":"{{ ui_service_name }}","verb":"get"}
 oauth_access_rule: "{{ oauth_default_openshift_sar if oauth_provider == 'openshift' }}"
 oauth_image_fqin: "{{ lookup('env', 'RELATED_IMAGE_OAUTH_PROXY') }}"
-oauth_ssl_port: 9443
+oauth_ssl_port: 8443
```
djzager, line 155: This is it? Why?

djzager (approved):
visack

roles/tackle/defaults/main.yml
jmontleon: I added a comment on the PR for main, apologies, I did not include it on this one:
`The external access network policy allows 8080 and 8443. So using 8443 instead of 9443 allows everything to work.`

I think my PR to add the oauth proxy predated the networkpolicy, so the auth proxy listening on 9443 at the time worked. That is no longer true.

djzager: heh. I saw it and didn't connect them.

PR #433, opened 2025-05-21
:ghost: Update path in case of downloaded binaries

jmontleon:


.github/workflows/create-release.yml
```diff
@@ -190,6 +190,7 @@ jobs:
         IMAGE_ORG="quay.io/konveyor"
         VERSION="${{ inputs.version }}"
         CSV="./bundle/manifests/konveyor-operator.clusterserviceversion.yaml"
+        export PATH=$PATH:./bin
```
aufi, line 193: Not sure about relative path in PATH, but worth trying, LGTM.

aufi (approved):

.github/workflows/create-release.yml
jmontleon: I tested quickly and it should work.

PR #437, opened 2025-06-02
:bug: Add requests and limits for additional containers

jmontleon:


roles/tackle/templates/customresource-rhbk-keycloak.yml.j2
```diff
@@ -17,6 +17,13 @@ spec:
       key: POSTGRES_PASSWORD
   proxy:
     headers: xforwarded
+  resources:
```
tmstff, line 20: When I tried it manually on my local machine, I used:

```
  keycloakDeploymentSpec:
    resources:
      requests:
        ...
```

tmstff: but the current version (of the branch) also seems to work fine, so maybe you can leave it as it is :-)

djzager (approved):

aufi (approved):

PR #495, opened 2025-11-13
:bug: Remove windup-shim from create-release changelog generation

jmontleon:
## Summary by CodeRabbit

* **Chores**
  * Updated release workflow configuration by removing an unused component from the batch release process.


jmontleon:
Maybe I am mistaken. I thought we stopped using it. Do we instead need to do something to generate a release for the repo so it doesn't stop up the release process when it tries to generate a changelog?

PR #499, opened 2025-11-19
:ghost: release-0.8 branch should be using release-0.8 images

shawn-hurley:
API tests PR: 371


shawn-hurley:
It appears as though we are getting significantly more tags than what the test is expecting. Can someone please look into this?

PR #531, opened 2026-02-02
:bug: Streamline RHBK deployment

jmontleon:
This should address https://issues.redhat.com/browse/MTA-6573

I think the 2-3 times is overblown, although I need to test more to be certain.

What I can definitively say is happening:
1. We create the RHBK CR
2. The RHBK statefulset starts rolling out. RHBK annotates missing secrets and hashes all secrets so it knows when it should restart to pick up changes. At this point we're pointing to a non existent TLS secret.
3. We annotate the RHBK service in order to create a cert secret
4. The SSL cert is created, the missing secret annotation is removed, the hash is updated, and as soon as RHBK becomes ready it is restarted to pick up changes

This change instead initially creates the RHBK CR with 0 instances to delay the statefulset rollout.
We then wait for the service to be created, annotate it, wait for the cert secret to be created, then update the CR to start the statefulset rollout.

## Summary by CodeRabbit

- New Features
  - Automatic detection and conditional creation of the RHBK Keycloak custom resource.
  - Waits for the RHBK service and TLS secret (with retries) before creating resources, and adds readiness checks.
  - Instance count for the Keycloak CR is now configurable instead of fixed.

- Bug Fixes
  - Reduces race conditions by ensuring TLS secret readiness precedes resource creation.

- Chores
  - Removed the database host entry from the Keycloak DB secret template.

### DIFF
--- a/roles/tackle/templates/secret-keycloak-db.yml.j2
+++ b/roles/tackle/templates/secret-keycloak-db.yml.j2
@@ -2,7 +2,6 @@ apiVersion: v1
 data:
   POSTGRES_DATABASE: {{ rhsso_db_name_b64 }} 
   POSTGRES_EXTERNAL_ADDRESS: {{ rhsso_db_host_b64 }}
-  POSTGRES_HOST:
   POSTGRES_PASSWORD: {{ rhsso_db_pass_b64 }}
   POSTGRES_USERNAME: {{ rhsso_db_user_b64 }}
   POSTGRES_SUPERUSER: {{ "true" | b64encode }}


roles/tackle/tasks/main.yml
```diff
@@ -512,6 +526,26 @@
...
+        api_version: v1
+        kind: Secret
+        name: "{{ rhbk_tls_secret_name }}"
+        namespace: "{{ app_namespace }}"
+      register: rhbk_secret
+      until: rhbk_secret.resources|length > 0
+      retries: 30
+      delay: 5
+
+    - name: "Set number of RHBK instances to 0 to prepare secret resources"
```
mguetta1: Looks like it should be `Set number of RHBK instances to 1`

roles/tackle/tasks/main.yml
```diff
@@ -512,6 +526,26 @@
...
+      set_fact:
+        rhbk_instances: 1
+
+    - name: "Create RHBK Keycloak CR"
```
mguetta1: *Update

roles/tackle/tasks/main.yml
jmontleon: Thank you for the corrections. See https://github.com/konveyor/operator/pull/534

PR #533, opened 2026-02-03
:bug: Fixes #472 Remove stateless kai resources when set to false

### LINKED ISSUE #472: 🐛 Setting experimental_deploy_kai: 'false' OR kai_solution_server_enabled: 'false' in the Tackle CR does not terminate the Kai solution server pods as expected.
Changing the experimental_deploy_kai flag in the Tackle Custom Resource to 'false' does not terminate or remove the Kai solution server pods as expected. The kai-api pods remain running in the openshift-mta namespace, indicating that the operator/controller did not reconcile the CR to disable the deployment.

**Steps :**

    1. Check existing Kai pods:
           ` kubectl -n openshift-mta get pods -l app=kai-api`
    2. Edit the Tackle CR to set experimental_deploy_kai: 'false'.
    3. Save the CR and wait for reconciliation.
    4. Observe that the kai-api pods are still running

**Expected Behavior:**

Setting experimental_deploy_kai: 'false' should cause the operator to remove all Kai-related pods and associated resources.

jmontleon:
## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured config always includes a proper models mapping and added a terminating newline to avoid parsing issues.

* **Refactor**
  * Simplified deployment logic to use centralized state flags, reducing nested conditions and improving reliability of component provisioning.

* **Chores**
  * Rendered database-related environment settings only when the corresponding service is enabled to prevent unnecessary configuration exposure.

### DIFF
--- a/roles/tackle/templates/kai/kai-api-deployment.yaml.j2
+++ b/roles/tackle/templates/kai/kai-api-deployment.yaml.j2
@@ -24,8 +24,10 @@ spec:
               value: /api
             - name: KAI_LLM_PARAMS
               value: '{{ kai_llm_params | to_json }}'
+{% if kai_solution_server_enabled|bool %}
             - name: KAI_DB_DSN
               value: 'postgresql+asyncpg://kai:{{ pg_password }}@{{ kai_database_address }}:5432/kai'
+{% endif %}
 {% if kai_api_key_secret_status.resources | length > 0 %}
 {% for (key, value) in kai_api_key_secret_status.resources.0.data.items() %}
             - name: {{ key }}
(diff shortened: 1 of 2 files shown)


roles/tackle/templates/kai/kai-api-deployment.yaml.j2
```diff
@@ -24,8 +24,10 @@ spec:
               value: /api
             - name: KAI_LLM_PARAMS
               value: '{{ kai_llm_params | to_json }}'
+{% if kai_solution_server_enabled|bool %}
```
fabianvf: Minor nit, this template is for the kai-solution-server, so I don't think we need this extra guard

fabianvf (approved):
Not sure why the test failed so unhelpfully

PR #578, opened 2026-06-10
:bug: Fix IdP CR upgrade issues.

### LINKED ISSUE #577: :bug: IdentityProvider URLs must be relative to the route.
Currently, when the operator builds the IdentityProvider for federation to the operator deployed Keycloak, it set's the issuer and redirectURI using service URLs. Instead they need to be relative to the route expressed in terms of the hub issuer.

jortel:
closes #577 

Also ensures only the IdentityProvider and Clients with the part-of label are manged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed API key preservation during upgrade scenarios where existing secrets are updated.

* **Chores**
  * Refactored identity provider client configuration management for improved maintainability.
  * Enhanced OIDC issuer URL configuration handling for better flexibility in different deployment environments.


roles/tackle/tasks/main.yml
```diff
@@ -73,10 +75,22 @@
...
+- name: "Check for existing IdentityProvider"
+  when:
+    - feature_auth_required|bool
+    - keycloak_detected|bool
+  k8s_info:
+    api_version: tackle.konveyor.io/v1alpha1
+    kind: IdentityProvider
+    name: keycloak
+    namespace: "{{ app_namespace }}"
+  register: idp_status
```
aufi, line 87: (got from claude, to be considered) The idp_status variable is only registered when feature_auth_required and keycloak_detected are both true (lines 78-87), but the subsequent task at line 93 references it unconditionally, causing a failure when those conditions aren't met. Add (idp_status is not defined) or before the existing condition on line 93 to handle this case safely.

aufi (approved):
Left one comment, overall LGTM.

roles/tackle/tasks/main.yml
jortel: I asked claude to review  the comment:

>  The review comment is incorrect. Look at the when conditions:
> 
>   - Line 78-81: Task "Check for existing IdentityProvider" runs when feature_auth_required|bool AND keycloak_detected|bool
>   - Line 90-93: Task "Create IdentityProvider" runs when feature_auth_required|bool AND keycloak_detected|bool AND (the idp_status check)
> 
>   Since line 93's condition includes the same two prerequisites (feature_auth_required and keycloak_detected), the task at line 89 will only execute when those conditions are met,
>   which means idp_status will always be defined when we reach line 93.
> 
>   The current code is correct - no fix needed. The when conditions on lines 90-92 ensure we only check idp_status when it has been registered.

aufi: Comment was incorrect, closing.

PR #593, opened 2026-08-25
:seedling: Update UI deployment to use HTTP GET liveness probe

### LINKED ISSUE #592: Update liveness probe on deployment to use httpGet instead of exec command
Update the liveness probe to work with the existing node/express server and any future changes to the UI container.

By using a httpGet probe, as is done in the `readinessProbe`, the operator will rely on the pod being able to listen and respond to an http request instead of if a the node process is running. Getting a response is a better choice as a node process could be running (or the grep node could be listed is the `ps -A` output), hung, and not actually be listening on the configured port.

sjd78:
This change updates the UI deployment to use an HTTP GET liveness probe instead of an exec probe. The exec probe could cause issues with the UI deployment if the UI deployment changes and no longer runs on node/express. The HTTP GET probe is a more reliable way to determine if the UI is running correctly.

Resolves: #592

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved UI deployment health checks by validating that the application responds successfully over HTTP.
  * Supports the configured UI protocol when performing liveness checks.

### DIFF
--- a/roles/tackle/templates/deployment-ui.yml.j2
+++ b/roles/tackle/templates/deployment-ui.yml.j2
@@ -84,9 +84,9 @@ spec:
               path: /
               port: {{ ui_port }}
               scheme: {{ ui_proto|upper }}
-            initialDelaySeconds: 10
-            timeoutSeconds: 1
-            periodSeconds: 5
+            initialDelaySeconds: 30
+            timeoutSeconds: 5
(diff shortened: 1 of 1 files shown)


djzager (approved):
ACK. Not sure why you would be having any issues with koncur tests.

fabianvf (approved):

PR #620, opened 2026-09-16
:sparkles: Restrict operand egress via NetworkPolicy

dymurray:
## What

The operator's operand NetworkPolicies (created at runtime, gated on `feature_isolate_namespace`, default `true`) previously restricted **ingress only** — egress was completely open. This adds egress restriction so operand traffic is governed in both directions.

Changes in `roles/tackle/templates/networkpolicy.yml.j2`:
- `*-deny-all` now lists `Egress` in `policyTypes` (default-deny egress as well as ingress).
- New `*-egress` policy (`podSelector: {}`) allows only:
  - **cluster DNS** — UDP/TCP 53
  - **intra-namespace** traffic — ui↔hub, hub→kai/llm-proxy, kai-api→kai-db, etc.
  - **outbound web + API server** — TCP 443/80 (analyzer package-repo pulls, external LLM providers) and 443/6443 (host-networked Kubernetes API server)

All other egress is blocked. Ingress policies are unchanged.

## Why

Implements the operand-egress portion of HPSTRAT-104 / **MTA-7874** (tailored operand NetworkPolicies restricting both ingress and egress; RH ProdSec / CIS 5.3.2 hardening). The RBAC to manage these policies already exists in the CSV.

## Notes
- Egress ships **on by default** (reuses `feature_isolate_namespace`); set `feature_isolate_namespace: false` on the Tackle CR to disable the policy set.
- The allow-list is deliberately pragmatic: analyzer/task pods (spawned by the Hub at runtime) egress to arbitrary Maven/package repos on 443/80, and KAI/llm-proxy reach external LLM providers on 443 — a blanket external allow on web ports keeps these working while blocking everything else.
- Policies remain runtime-applied (not added to `bundle/`), consistent with ISV-6226.

## Testing
- Template renders to valid YAML (5 NetworkPolicy objects; `deny-all` = Ingress+Egress, `egress` = 3 allow rules).
- Reviewer e2e: with the CR applied, `oc get networkpolicy -n <ns>` shows the egress policy; analysis jobs complete, UI/hub reachable, metrics scrape succeeds.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Security**
  * Outbound traffic remains denied by default for application pods, with explicit exceptions for DNS, same-namespace communication, HTTP/HTTPS traffic, and Kubernetes API access.
  * DNS access is scoped to cluster DNS namespaces, and API connectivity on TCP port 8443 is supported.
  * Network policy ordering helps preserve required egress access during reconciliation.

* **Documentation**
  * Added guidance on namespace isolation, restricted outbound ports, and configuring supplemental policies for additional external services.

* **Bug Fixes**
  * Disabling namespace isolation now removes policies previously managed by the operator.


kaovilai (commented):
what about this?

Setting feature_isolate_namespace: false currently skips reconciliation but leaves previously created policies active. Could we explicitly remove the operator-managed policies when isolation is disabled? This also avoids relying on multi-document template deletion behavior.
```yaml
- name: "Create Network Policies"
  k8s:
    state: present
    definition: "{{ lookup('template', 'networkpolicy.yml.j2') }}"
  when: feature_isolate_namespace | bool

- name: "Remove Network Policies when namespace isolation is disabled"
  k8s:
    state: absent
    api_version: networking.k8s.io/v1
    kind: NetworkPolicy
    namespace: "{{ app_namespace }}"
    name: "{{ item }}"
  loop:
    - "{{ app_name }}-namespace"
    - "{{ app_name }}-external"
    - "{{ app_name }}-metrics"
    - "{{ app_name }}-egress"
    - "{{ app_name }}-deny-all"
  when: not (feature_isolate_namespace | bool)
```

jmontleon (approved):
ACK

dymurray:
@kaovilai good catch — this is a real reversibility bug. As written, flipping `feature_isolate_namespace` off just skipped the reconcile and left the default-deny egress live, so "disable isolation" didn't actually restore open networking. Implemented your `state: absent` removal loop in 496c9b0. I kept the explicit name list rather than deleting by the `app=` label selector, since a label-based delete could catch NetworkPolicies this operator doesn't own. Thanks!

roles/tackle/templates/networkpolicy.yml.j2
```diff
@@ -60,3 +48,60 @@ spec:
...
+  # ClusterIP such as the DNS service, so the DNS namespace is selected with no
+  # ports rather than `ports: 53`. Both the OpenShift (openshift-dns) and
+  # upstream Kubernetes (kube-system) DNS namespaces are allowed.
+  - to:
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: openshift-dns
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: kube-system
```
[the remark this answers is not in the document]
dymurray: Acknowledged, but declining for now. Scoping the port-less DNS rule to CoreDNS pods is correct in the abstract, but the benefit is marginal here — every operand already has broad egress to the internet on 443/80, so lateral access to a kube-system pod on an odd port isn't the meaningful exposure (and those pods are largely host-networked and gated by their own ingress). Against that, `namespaceSelector`+`podSelector` is AND-matched and the DNS pod label differs by platform (`dns-default` on OpenShift vs `k8s-app: kube-dns` upstream/minikube), so a wrong label silently breaks DNS on a whole platform — a failure mode this PR has already had to fix twice. If pod-level DNS scoping becomes a requirement, we'll do it as a follow-up validated on both OpenShift and minikube.

roles/tackle/templates/networkpolicy.yml.j2
```diff
@@ -60,3 +61,36 @@ spec:
...
+    - protocol: TCP
+      port: 53
+  # Intra-namespace traffic (ui<->hub, hub->kai/llm-proxy, kai-api->kai-db, etc.).
+  - to:
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: {{ app_namespace }}
+  # Outbound web + Kubernetes API server. 443/80 cover analyzer package-repo
+  # pulls and external LLM providers; 443/6443 cover the host-networked API server.
+  - ports:
```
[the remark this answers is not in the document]
dymurray: Declining. The API server is host-networked, so it's in no namespace and `namespaceSelector` can't target it — leaving only `to.ipBlock` with node CIDRs, which you correctly note can't be the pre-DNAT ClusterIP. But node IPs are dynamic (scaling, replacement, multiple control-plane nodes, LB front-ends), so a pinned CIDR that drifts would silently sever every operand's API access and CrashLoop them — which is exactly the outage this PR's 8443 commit (2ad5a77) just fixed. The security gain is negligible since 443 is already open cluster-wide, and 6443/8443 are near-exclusively kube API-server ports that still require the destination's own ingress to accept traffic. The port-only rule is the standard pragmatic pattern precisely because there's no portable destination selector for a host-networked API server.

roles/tackle/templates/networkpolicy.yml.j2
```diff
@@ -60,3 +48,65 @@ spec:
...
+  # 443 (managed Kubernetes), 6443 (OpenShift), 8443 (minikube/k3s-style).
+  - ports:
+    - protocol: TCP
+      port: 443
+    - protocol: TCP
+      port: 80
+    - protocol: TCP
+      port: 6443
+    - protocol: TCP
+      port: 8443
```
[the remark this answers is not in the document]
dymurray: Declining. The API server is host-networked, so it's in no namespace and `namespaceSelector` can't target it — leaving only `to.ipBlock` with node CIDRs, which you correctly note can't be the pre-DNAT ClusterIP. But node IPs are dynamic (scaling, replacement, multiple control-plane nodes, LB front-ends), so a pinned CIDR that drifts would silently sever every operand's API access and CrashLoop them — which is exactly the outage this PR's 8443 commit (2ad5a77) just fixed. The security gain is negligible since 443 is already open cluster-wide, and 6443/8443 are near-exclusively kube API-server ports that still require the destination's own ingress to accept traffic. The port-only rule is the standard pragmatic pattern precisely because there's no portable destination selector for a host-networked API server.

kaovilai (approved):
Ew opus 4.8

jortel (approved):
LGTM

roles/tackle/tasks/main.yml
```diff
@@ -359,6 +359,21 @@
...
+    api_version: networking.k8s.io/v1
+    kind: NetworkPolicy
+    namespace: "{{ app_namespace }}"
+    name: "{{ item }}"
+  loop:
+    - "{{ app_name }}-namespace"
+    - "{{ app_name }}-external"
+    - "{{ app_name }}-metrics"
+    - "{{ app_name }}-egress"
+    - "{{ app_name }}-deny-all"
```
fabianvf, line 374: Is this ordering something we have to worry about during deletion? It's a695e1e in reverse - `-egress` goes at 373 and `-deny-all` at 374, so there's a window with default-deny egress and no allow rules. Would we break the operator mid-cleanup?

roles/tackle/tasks/main.yml
```diff
@@ -359,6 +359,21 @@
...
+    state: absent
+    api_version: networking.k8s.io/v1
+    kind: NetworkPolicy
+    namespace: "{{ app_namespace }}"
+    name: "{{ item }}"
+  loop:
+    - "{{ app_name }}-namespace"
+    - "{{ app_name }}-external"
+    - "{{ app_name }}-metrics"
+    - "{{ app_name }}-egress"
```
fabianvf, line 373: Related but without the flag: delete just `-egress` on a live install and `-deny-all` still denies egress, so I don't think the operator can reach the API to put it back. Could add a finalizer on `-egress`, or drop `Egress` from `-deny-all` and let `-egress` be the whole egress policy

roles/tackle/templates/networkpolicy.yml.j2
```diff
@@ -60,3 +48,65 @@ spec:
...
+  - to:
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: {{ app_namespace }}
+  # Outbound web + Kubernetes API server. 443/80 cover analyzer package-repo
+  # pulls and external LLM providers. The host-networked API server is reached
+  # via a ClusterIP that DNATs to the node IP, and the CNI matches egress against
+  # that post-DNAT node IP + port, so the actual API server port must be allowed:
+  # 443 (managed Kubernetes), 6443 (OpenShift), 8443 (minikube/k3s-style).
+  - ports:
```
fabianvf, line 86: Do we need to cover the hub's Proxy setting here, and repo URLs on a nonstandard port? `Proxy` takes a free-form port (`shared/api/core.go:108`) and git clones whatever port is in the URL, so both land outside 80/443. I don't know how much either gets used in the field, is that a supported config or do we not care?

roles/tackle/tasks/main.yml
dymurray: Good catch — you're right that it was a695e1e in reverse. Rather than reorder the deletion, I fixed the root cause: `-deny-all` no longer denies egress at all (see eec77a6). The `-egress` policy already isolates every pod for egress (`podSelector: {}` + `policyTypes: [Egress]`) and carries the allow-list, so it's a complete default-deny-egress on its own — the `Egress` entry in `-deny-all` was redundant. With it gone, deleting `-egress` reopens egress immediately, so there's no longer any window of default-deny-egress-with-no-allows mid-cleanup regardless of order.

roles/tackle/tasks/main.yml
dymurray: Exactly right, and this is the more important of the two — thanks. I went with your second suggestion (drop `Egress` from `-deny-all`, let `-egress` be the whole egress policy) in eec77a6, since it's simpler than a finalizer and fixes both this and the teardown-ordering window above. Now if `-egress` is deleted or lost on a live install, nothing else keeps denying egress, so the operator can reach the API server and recreate it instead of deadlocking. Steady-state egress posture is unchanged because `-egress` itself is untouched.

roles/tackle/templates/networkpolicy.yml.j2
dymurray: Valid — I confirmed `Proxy.Port` is a free-form `int` (shared/api/core.go:114) and git clones whatever port the repo URL carries, so both can land outside 80/443. These are supported configs, but the ports are user-defined and can't be enumerated in a static allow-list. I've documented them explicitly in the README (d6e489c/eec77a6): the hub Proxy on a non-standard port and git repo URLs on non-standard ports are now called out as first-class cases that need a supplemental, user-managed NetworkPolicy, with a copy-paste example. If we find these are common enough in the field that the friction outweighs the isolation benefit, the fallback is `feature_isolate_namespace: false` — but I'd rather not widen the default allow-list to arbitrary ports. Open to thoughts.

fabianvf: Well the hub has kube access, so the other option would be having it enumerate the specific ports somewhere and either have it manage some policy or have the operator do it. Or at least read the networkpolicy and let users know that the configuration will be broken unless an administrator updates the network policy.

shubham-pampattiwar (approved):

roles/tackle/templates/networkpolicy.yml.j2
dymurray: based on the docs I read its better to force the user to explicitly need to update the network policy for our use case rather than auto updating the egress.

And its unclear with a cluster wide proxy if our own existing egress rules woudln't be good enough there

